PR #185, opened 2014-12-04
ppsspp 0.9.9.1: new formula

tomyun:
[PPSSPP](http://ppsspp.org) is an open source emulator of Sony PlayStation Portable (PSP).

Here are some notes worth mentioning:

1. Their source archive couldn't be built without bringing in some dependencies from git repositories. That's why it's directly tapping git tag which spends quite a long time to clone large submodules including its own `ffmpeg`.

2. Even though `ffmpeg` submodule is fully cloned, it's not compiled anyways, but only pre-built libraries are seemingly used on linkage. A patch would be needed if we really want to build everything from source.

3. The build system looks for a version with `git distribute` which is not available on staging step. As a workaround, its version is manually set to the one of formula being installed.

4. `-DCMAKE_BUILD_TYPE=None` had to be removed from `std_cmake_args`, otherwise it won't build. The build system assumes `Release` mode when no build type is specified.

5. Final product is an `.app` package. A symlink is provided for running on command line.


mistydemeo:
Ping @mikemcquaid - do you know why bottles aren't being uploaded here? It built fine, and the only CI failure is the lack of the `test do` block.

mikemcquaid:
The Bottles job never finished: http://bot.brew.sh/job/Homebrew%20Games%20Pull%20Requests/78/downstreambuildview/

You can retry the main job.

mikemcquaid:
This is the new job: http://bot.brew.sh/job/Homebrew%20Games%20Pull%20Requests/85/downstreambuildview/

mikemcquaid:
Bottles uploaded now: http://bot.brew.sh/job/Homebrew%20Bottles/12513/console

PR #184, opened 2014-12-04
mame: disable LTO on OS X 10.7 and below

tomyun:
Recent addition of LTO (link time optimization) in the formula seems to be broken on older OS X. It builds without problem, but crashes on startup at least with my 10.7 (Lion). An exactly same issue found with 0.155 formula as well.

It's probably due to an obsolete version of clang which is no longer updated/supported as Xcode and OS itself are. As an workaround, the formula is updated to ensure that LTO is only enabled on 10.8 (Mountain Lion) and above.

FYI, here is a part of crash log from mame 0.156 with SDL2:

```
Crashed Thread:  0  Dispatch queue: com.apple.main-thread

Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: 0x000000000000000d, 0x0000000000000000

VM Regions Near 0:
-->
    __TEXT                 00000001072ef000-000000010ac53000 [ 57.4M] r-x/rwx SM=COW  /usr/local/Cellar/mame/0.156/bin/mame

Application Specific Information:
objc[5802]: garbage collection is OFF

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libunwind.dylib                    0x00007fff8fd86fe6 libunwind::CompactUnwinder_x86_64<libunwind::LocalAddressSpace>::stepWithCompactEncodingFrameless(unsigned int, unsigned long long, libunwind::LocalAddressSpace&, libunwind::Registers_x86_64&, bool) + 630
1   libunwind.dylib                    0x00007fff8fd8520c libunwind::CompactUnwinder_x86_64<libunwind::LocalAddressSpace>::stepWithCompactEncoding(unsigned int, unsigned long long, libunwind::LocalAddressSpace&, libunwind::Registers_x86_64&) + 52
2   libunwind.dylib                    0x00007fff8fd85171 libunwind::UnwindCursor<libunwind::LocalAddressSpace, libunwind::Registers_x86_64>::step() + 69
3   libobjc.A.dylib                    0x00007fff8e9b9294 objc_addExceptionHandler + 82
4   com.apple.CoreFoundation           0x00007fff91f1ca34 _CFDoExceptionOperation + 308
5   com.apple.AppKit                   0x00007fff8ad21696 _NSAppKitLock + 73
6   com.apple.AppKit                   0x00007fff8af1c0c3 +[NSScroller initialize] + 95
7   libobjc.A.dylib                    0x00007fff8e9b4662 _class_initialize + 320
[...]

PR #201, opened 2015-01-01
mamedev 0.157

mbcoguno:
* Style updates.
* mame, mess, rom-tools: add manpages.
* mess, ume: enable LTO if compiler support.

### DIFF
--- a/mame.rb
+++ b/mame.rb
@@ -1,10 +1,8 @@
-require "formula"
-
 class Mame < Formula
   homepage "http://mamedev.org/"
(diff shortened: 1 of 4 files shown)


mbcoguno:
BTW. since #184 isn't accept or reject, I just simply add LTO support in mess and ume like mame. I will update them once #184 is closed.

Another question, mame can use external library instead itself, should be add optional dependency?

mistydemeo:
Thanks!

PR #202, opened 2015-01-05
residualvm 0.2.1

steinuil:
Updated ResidualVM to 0.2.1.


jacknagel:
@BrewTestBot test this please

steinuil:
Fixed the checksum warning and rebased. By the way, is it better practice to use sha1 or sha256?

PR #208, opened 2015-01-24
fuego: update revision 1981

milescrabill:
older versions of fuego don't work with recent boost versions
revision 1981 is the current as of this pull request, but realistically this formula should be head only until a stable version is determined

### DIFF
--- a/fuego.rb
+++ b/fuego.rb
@@ -1,14 +1,16 @@
-require 'formula'
-
 class Fuego < Formula
-  url 'https://downloads.sourceforge.net/project/fuego/fuego/1.0/fuego-1.0.tar.gz'
   homepage 'http://fuego.sourceforge.net/'
-  sha1 '1fe3de726d7020278f7cb2f678a1909053bf0107'
+  head "http://svn.code.sf.net/p/fuego/code/trunk"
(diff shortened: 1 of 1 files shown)


mistydemeo:
Do you know if a stable version is planned?

milescrabill:
I haven't gotten in contact with the maintainer, but the latest tar is dated 2013-11-12, and the [recommended install instructions](http://fuego.sourceforge.net/mac-install.html) say to install this way.

fuego.rb
```diff
@@ -1,14 +1,16 @@
...
-
 class Fuego < Formula
-  url 'https://downloads.sourceforge.net/project/fuego/fuego/1.0/fuego-1.0.tar.gz'
   homepage 'http://fuego.sourceforge.net/'
-  sha1 '1fe3de726d7020278f7cb2f678a1909053bf0107'
+  head "http://svn.code.sf.net/p/fuego/code/trunk"
+  url "http://svn.code.sf.net/p/fuego/code/trunk", :revision => 1981
+  version "1.1.SVN"
 
+  # automake should be a build dep, but won't install otherwise
```
jacknagel: There's no practical difference between a regular dep (or one marked explicitly with `=> :run`) and a `=> :build` dep in the build environment, so I'm not sure what this means.

PR #210, opened 2015-02-02
ppsspp 1.0

tomyun:
1.0 version is released.

- Use `sdl2` replacing `sdl`.
- FFmpeg patch removed (hrydgard/ppsspp#5772)
- A fake test always returning true added for feeding test-bot.
- Regex syntax changed according to `brew audit --strict`.


jacknagel:
@BrewTestBot test this please

PR #211, opened 2015-02-09
mapcrafter 1.5.4 (new formula)

BM5k:


mapcrafter.rb
```diff
@@ -0,0 +1,24 @@
...
+  sha1 "ffe7d17a57099506b572bb6866ea76c8ad5f0564"
+
+  depends_on "boost"
+  depends_on "cmake" => :build
+  depends_on "jpeg-turbo"
+  depends_on "libpng"
+
+  def install
+    args = std_cmake_args
+    args << "-DJPEG_INCLUDE_DIR=/usr/local/opt/jpeg-turbo/include/"
```
mistydemeo: Can't hardcode these paths. You can get the location from creating a `Formula` object:

```ruby
args << "-DJPEG_INCLUDE_DIR=#{Formula["jpeg-turbo"].opt_include}"
```

mapcrafter.rb
```diff
@@ -0,0 +1,24 @@
...
+    args << "-DJPEG_LIBRARY=/usr/local/opt/jpeg-turbo/lib/libjpeg.dylib"
```
mistydemeo: Similarly, here too.

mapcrafter.rb
```diff
@@ -0,0 +1,24 @@
+class Mapcrafter < Formula
+  homepage "http://mapcrafter.org"
+  url "https://github.com/mapcrafter/mapcrafter/archive/v.1.5.4.tar.gz"
+  sha1 "ffe7d17a57099506b572bb6866ea76c8ad5f0564"
+
+  depends_on "boost"
+  depends_on "cmake" => :build
```
mistydemeo: Can you place this dep at the top of the list?

mapcrafter.rb
BM5k: thanks, I just pulled these lines from the official [installation instructions][1], I’ll update them to be dynamic.

[1]: https://github.com/mapcrafter/mapcrafter/blob/c479009ae344863ef61151d7eb8a8fa8ae9b5953/docs/installation.rst

mapcrafter.rb
BM5k: @mistydemeo I’m not sure how to change line #14, is this correct?

`"-DJPEG_LIBRARY=#{Formula["jpeg-turbo”].lib}/libjpeg.dylib"`

mapcrafter.rb
BM5k: of course!

mapcrafter.rb
mistydemeo: Yep, though `opt_lib`. Thanks!

BM5k:
@mistydemeo how’s that?

mistydemeo:
Looks good!

Build is failing on Mt. Lion:

```
/tmp/mapcrafter-5ZUJjs/mapcrafter-v.1.5.4/src/mapcraftercore/config/../util/../compat/thread.h:29:12: fatal error: 'condition_variable' file not found
#  include <condition_variable>
           ^
```

BM5k:
@mistydemeo any suggestions for how to troubleshoot this?

This version may not be supported on Mt. Lion? I’ll check with the mapcrafter project.

mistydemeo:
Do you know if mapcrafter requires C++11?

BM5k:
I don't, but I can try to find out

BM5k:
@mistydemeo yeah, it looks like they do require C++11.

mistydemeo:
Okay - you can add `needs :cxx11` above the dependency list and `ENV.cxx11` at the top of the install method.

The boost dep will also need to be replaced with this:

```ruby
  if MacOS.version < :mavericks
    depends_on "boost" => "c++11"
  else
    depends_on "boost"
  end
```

mistydemeo:
Thanks very much! Squashed to one commit and merged.

BM5k:
@mistydemeo thanks for your help!

PR #212, opened 2015-02-09
Added caveat and retina display patch dwarf-fortress

noqqe:
Out of the box, after a brew install, you have to do some things to make dwarf fortress work. For new users, it should be easier now.


dwarf-fortress.rb
```diff
@@ -12,4 +12,26 @@ def install
     rm_rf "sdl" # only contains a readme
     libexec.install Dir["*"]
   end
+  
+  def caveats; <<-EOS.undent
+    DwarfFortress heavily relies on X11. After you installed Xquartz
+    create a symlink to match the LibraryPath of DwarfFortress.
+    Do:
+      sudo ln -s /usr/X11 /usr/X11R6
```
mistydemeo: I don't want to suggest that users make this change to their system. Instead, it would be better to use `install_name_tool` to point the library at a different location.

What version of Dwarf Fortress are you encountering this with, however? The linkage in 0.40.23 is to `/usr/X11`, not `/usr/X11R6`.

dwarf-fortress.rb
```diff
@@ -12,4 +12,26 @@ def install
...
 end
+__END__
+--- a/data/init/init.txt
++++ b/data/init/init.txt
+@@ -64,7 +64,7 @@ On a multi-core machine none of this is very likely to matter; stick to 2D.
+
+ Linux/OS X users may also use PRINT_MODE:TEXT for primitive ncurses output.
+
+-[PRINT_MODE:2D]
++[PRINT_MODE:STANDARD]
```
mistydemeo: What does this change mean to users who aren't on Retina Macs?

dwarf-fortress.rb
mistydemeo: `otool -L` output for reference:

```
otool -L /usr/local/Cellar/dwarf-fortress/0.40.23/libexec/libs/SDL_ttf.framework/SDL_ttf
/usr/local/Cellar/dwarf-fortress/0.40.23/libexec/libs/SDL_ttf.framework/SDL_ttf:
	/usr/local/opt/dwarf-fortress/libexec/libs/SDL_ttf.framework/Versions/A/SDL_ttf (compatibility version 1.0.0, current version 1.0.0)
	/usr/X11/lib/libfreetype.6.dylib (compatibility version 10.0.0, current version 10.20.0)
	@executable_path/../Frameworks/SDL.framework/Versions/A/SDL (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 123.0.0)
```

dwarf-fortress.rb
noqqe: Standardmode also works for non Retina Displays - Just a different way of game resolution. Its a config file, so you can change that value. But out of the box for most of the users (with retina) it will not work at all.

dwarf-fortress.rb
noqqe: Okay, I can completly understand that. you are right.
I deleted my symlink again and still recieve this error 

```
dyld: Library not loaded: /usr/X11R6/lib/libfreetype.6.dylib
  Referenced from: /usr/local/Cellar/dwarf-fortress/0.40.23/libexec/libs/SDL_ttf.framework/Versions/A/SDL_ttf
  Reason: no suitable image found.  Did find:
	/usr/local/lib/libfreetype.6.dylib: mach-o, but wrong architecture
```

mistydemeo: What is the output of `ls -l /usr/X11/lib/libfreetype.6.dylib` for you?

noqqe:
```
$ ls -l /usr/X11/lib/libfreetype.6.dylib
-rwxr-xr-x  1 root  wheel  1143056 Aug 12  2014 /usr/X11/lib/libfreetype.6.dylib
$
```

mistydemeo: What is the output of `otool -L /usr/local/Cellar/dwarf-fortress/0.40.23/libexec/libs/SDL_ttf.framework/Versions/A/SDL_ttf`?

I'm unable to reproduce your error even if I remove `/usr/X11R6`. Does the problem recur if you `brew reinstall dwarf-fortress`? (Make sure you back up your saves first!)

noqqe: screen shot 2015-02-15 at 08 37 15

mistydemeo:
Merged alternate caveats, thanks!

PR #223, opened 2015-03-11
Update cataclysm to version 0.C

sparr:


mistydemeo:
Thanks very much!

(For future reference, our preferred commit message format for version bumps is `formula version`, e.g. `cataclysm 0.C`.)

PR #224, opened 2015-03-11
wesnoth 1.12.1

dunn:
The test might have to be removed; on my system I have a full-screen terminal and it jumps away when I run the test, as if it's going to open the windowed application.  If that happens for other people and if it's unacceptable behavior from a test I'll yank it.


mistydemeo:
Passes on the bot, which is good enough for me! Thanks!

PR #225, opened 2015-03-13
More compiling options for minetest

neoascetic:


minetest.rb
```diff
@@ -47,10 +45,10 @@ def install
     (buildpath/"games/minetest_game").install resource("minetest_game")
 
     args = std_cmake_args - %w{-DCMAKE_BUILD_TYPE=None}
-    args << '-DCMAKE_BUILD_TYPE=Release'
-    args << "-DENABLE_FREETYPE=1" if build.with? "freetype"
-    args << "-DBUILD_CLIENT=0" if build.without? "client"
-    args << "-DBUILD_SERVER=0" if build.without? "server"
+    args << "-DCMAKE_BUILD_TYPE=Release" << "-DBUILD_CLIENT=1" << "-DBUILD_SERVER=0"
```
mistydemeo: What's the reason to remove these options?

neoascetic: In result of compiling bundle is built. Separate server executable is not required, but client always required since bundle won't work otherwise.

minetest.rb
```diff
@@ -47,10 +45,10 @@ def install
...
+    args << "-DENABLE_REDIS=1" if build.with? "redis"
+    args << "-DENABLE_LEVELDB=1" if build.with? "leveldb"
+    args << "-DENABLE_FREETYPE=1" << "-DCMAKE_EXE_LINKER_FLAGS='-L/usr/local/lib'" if build.with? "freetype"
```
neoascetic: @mistydemeo can I use some getter here instead of hardcoded `/usr/local/lib`?

neoascetic:
@mistydemeo any update?

minetest.rb
mistydemeo: Sorry for the late reply. Yes, you can do `-L#{Formula["freetype"].opt_lib}`.

neoascetic:
PR was updated

mistydemeo:
Pulled, thanks!

PR #227, opened 2015-03-16
New formula: openclonk

Meowtimer:
Adds a formula for the game openclonk - http://www.openclonk.org


openclonk.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on "freetype"
+  depends_on "glew"
+  depends_on "libogg"
+  depends_on "libvorbis"
+  depends_on "boost"
+  depends_on "freealut"
+
+  def install
+    args = std_cmake_args
+    if Formula["freetype"].installed?
```
mistydemeo: Since freetype is an unconditional dependency, this is always true - you can remove the if statement.

openclonk.rb
```diff
@@ -0,0 +1,32 @@
...
+      args << "-DFREETYPE_INCLUDE_DIR_freetype2=#{HOMEBREW_PREFIX}/include/freetype2"
```
mistydemeo: Instead of `HOMEBREW_PREFIX/include`, use paths from in freetype's prefix:

```ruby
freetype = Formula["freetype"]
args << "-DFREETYPE_INCLUDE_DIR_freetype2=#{freetype.include}/freetype2"
args << "-DFREETYPE_INCLUDE_DIR_ft2build=#{freetype.include}/freetype2"
```

openclonk.rb
Meowtimer: Oh, whoops. Right

mistydemeo:
Looks like this requires C++11 - can you add `needs :cxx11` below the deps and `ENV.cxx11` at the top of the install method?

mistydemeo:
@BrewTestBot test this please

mistydemeo:
Looks like this is still failing on Mt. Lion, for new reasons:

```
/tmp/openclonk20150425-84480-rl8m25/openclonk-6.0/src/platform/StdSchedulerMac.mm:222:10: error: no visible @interface for 'NSTimer' declares the selector 'setTolerance:'
                [timer setTolerance:0.0];
                 ~~~~~ ^~~~~~~~~~~~
```

mistydemeo:
Looks like the `setTolerance:` method was introduced in 10.9.

mistydemeo:
Submitted a patch for this to the developers: openclonk/openclonk#6

mistydemeo:
The patch has been accepted! Mind incorporating it into the formula?

Meowtimer:
Thanks! Now I'm not sure what to point the formula's url to, though. The archive from the pull request commit would represent a development snapshot. Maybe wait for openclonk 6.1? That is somewhat around the corner, I think.

mistydemeo:
Could either wait for 6.1, or apply it as a `patch do` block to the existing stable version.

Meowtimer:
I guess I have to do something like http://stackoverflow.com/questions/11707936/c-mac-preprocessor-flag-for-base-sdk for occlusionState usage

mistydemeo:
Thanks! Pulled with a few tweaks.

PR #233, opened 2015-04-03
snescom 1.7.4.1

dholm:


snescom.rb
```diff
@@ -0,0 +1,30 @@
+class Snescom < Formula
+  homepage "http://bisqwit.iki.fi/source/snescom.html"
+  url "http://bisqwit.iki.fi/src/arch/snescom-1.7.4.1.tar.bz2"
+  sha256 "e80187d969fd448fe4d3c636bdbe9c24fe452e1436566b29fbf3410bde739504"
+
+  patch :p1 do
+    # Includes cstdlib for Clang.
```
mistydemeo: Can you report the need for this patch to the developers, please? Thanks!

snescom.rb
```diff
@@ -0,0 +1,30 @@
...
+    url "https://gist.githubusercontent.com/protomouse/31335db2fe5c11978faf/raw/f0671999314e87c0487035e4f80f92a67a0c4d06/snescom-disasm-missing-cstdlib.diff"
+    sha256 "3e214ae0a5751dfe72f81c0723b792601ddb4bca335dd2d5fd345572f051e29c"
+  end
+
+  depends_on "boost"
```
mistydemeo: I don't see dynamic linkage against boost - seems this is only necessary at buildtime. This can be:

```ruby
depends_on "boost" => :build
```

dholm: Fixed, thanks!

snescom.rb
dholm: Done!

mistydemeo: Thanks! Is there a public mailing list/issue tracker you can link to here? Otherwise, you can just add a comment saying that it was reported.

dholm: I sent the author an email.  His reply was:

    "I will add this fix to the next release. Thanks for the report."

mistydemeo:
Thanks! Merged, with a comment added to that patch.

PR #234, opened 2015-04-05
Add ckan.rb v1.6.6

NathanJang:
This is [The Comprehensive Kerbal Archive Network (CKAN)](https://github.com/KSP-CKAN/CKAN).

A short wrapper script for it is hardcoded as a heredoc. Please advise if there is a better method to specify a wrapper script.


ckan.rb
```diff
@@ -0,0 +1,24 @@
+class Ckan < Formula
+  homepage "https://github.com/KSP-CKAN/CKAN/"
+  version 'v1.6.6'
+  # CKAN releases a raw .exe file
+  url "https://github.com/KSP-CKAN/CKAN/releases/download/#{version}/ckan.exe", :using => :nounzip
```
mistydemeo: Just inline the version into the URL, please.

ckan.rb
```diff
@@ -0,0 +1,24 @@
+class Ckan < Formula
+  homepage "https://github.com/KSP-CKAN/CKAN/"
+  version 'v1.6.6'
```
mistydemeo: Nitpick, but use double-quotes everywhere, please.

mistydemeo:
Is it possible to write a test for this?

NathanJang:
Is this better?

ckan.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    # The usual method for running ckan.exe is `mono ckan.exe arg1 arg2 ...`
+    # This is a wrapper script for aliasing it as `ckan arg1 arg2 ...` into the PATH
+    wrapper_script = <<-EOS.undent
+      #!/bin/bash
+      mono #{prefix}/ckan.exe $@
+    EOS
+
+    File.write("ckan", wrapper_script)
+    bin.install "ckan"
```
xu-cheng:
```ruby
  def install
    (libexec/"bin").install "ckan.exe"
    (bin/"ckan").write <<-EOS.undent
      #!/bin/sh
      exec mono "#{libexec}/bin/ckan.exe" "$@"
    EOS
  end
```

xu-cheng:
You can add below code in the beginning and use `depends_on MonoDependency `. So users don't have to install homebrew mono.
```ruby
class MonoDependency < Requirement
  fatal true
  default_formula "mono"
  satisfy { which("mono") }
end
```

NathanJang:
Thanks for the suggestions @xu-cheng!

ckan.rb
```diff
@@ -0,0 +1,29 @@
...
+    # The usual method for running ckan.exe is `mono ckan.exe arg1 arg2 ...`
+    # This is a wrapper script for aliasing it as `ckan arg1 arg2 ...` into the PATH
+    (bin/"ckan").write <<-EOS.undent
+      #!/bin/sh
+      exec mono "#{libexec}/bin/ckan.exe" "$@"
+    EOS
+  end
+
+  test do
+    system "ckan", "version"
```
xu-cheng: Sorry, just found out this. You need to use full path name. And it's better to have a little complex test. Like this:
```
system bin/"ckan", "available"
```

NathanJang:
Okay.

NathanJang:
The problem is that CKAN is quite buggy and I'm not sure whether this will pass with every update.

xu-cheng:
Ok, it seems `available` requires KSP installed. You can switch back to `system bin/"ckan", "version"`.

xu-cheng:

NathanJang:
Welcome!

NathanJang:
@xu-cheng I have since learned that CKAN requires MDK and not MRE. The homebrew mono is inadequate. Is there any way to declare such a dependency? If it helps, MDK can be installed through homebrew cask.

xu-cheng:
I think MRE is enough. However, the homebrew mono's GUI is broken. (See https://github.com/Homebrew/homebrew/issues/35364). So you can only use CKAN in CLI mode if you are using homebrew mono. 

You can add caveats to inform users to install MRE from cask if they want GUI CKAN.

NathanJang:
It appears that the cask for MRE is not referencing a valid URL anymore. If you've installed MRE, could you point me to the download URL? I can't seem to find it on the official website.

PR #236, opened 2015-04-07
angband 3.5.1

dunn:
Got in a bit over my head here, and I don't think I did the ncurses stuff right.  Would using `homebrew/dupes/ncurses` be better?


dunn:
Updated; probably the Wrong way to detect ncurses-config in a version-agnostic way, but probably better than hardcoding `ncurses5.4-config`?

angband.rb
```diff
@@ -1,20 +1,38 @@
...
+  depends_on "homebrew/dupes/tcl-tk" => "with-x11" if build.with? "x11"
 
   def install
-    system "./configure", "--prefix=#{prefix}",
-                          "--enable-curses",
-                          "--disable-x11",
-                          "--disable-sdl",
-                          "--disable-sdl-mixer",
-                          "--with-libpath=#{libexec}"
+    ENV["NCURSES_CONFIG"] = `ls #{MacOS.sdk_path}/usr/bin/ncurses*config`.chomp
```
mistydemeo: Apple has shipped ncurses 5.4 since OS X 10.4, so it should be safe just to hardcode `5.4` as the version.

dunn:
Back to `ncurses5.4-config` and found a mirror also.

mistydemeo:
Thanks!

PR #241, opened 2015-04-22
Updated CKAN to 1.6.10

aw94:


xu-cheng:

PR #242, opened 2015-04-23
gnu-chess 6.2.1

dunn:
Made the "book" optional on account of its size; or should it be recommended?


mistydemeo:
> should it be recommended?

I honestly have no idea, so let's go with this and see if any users complain! Thanks!

PR #243, opened 2015-04-23
frobtads: update to new upstream HEAD repo

ziz:
Link from upstream homepage http://www.tads.org/frobtads.htm


xu-cheng:

PR #251, opened 2015-05-03
nazghul 0.7.1

dunn:
Is there no way to make formula aliases outside of Homebrew core?


nazghul.rb
```diff
@@ -1,27 +1,34 @@
-require 'formula'
-
 class Nazghul < Formula
-  url 'https://downloads.sourceforge.net/project/nazghul/nazghul/nazghul-0.7.0/nazghul-0.7.0.tar.gz'
-  homepage 'http://myweb.cableone.net/gmcnutt/nazghul.html'
-  sha1 '1d67c3fb3b03b3f8b5170305f3d97428932e6c4f'
+  homepage "https://web.archive.org/web/20130402222926/http://myweb.cableone.net/gmcnutt/nazghul.html"
```
dunn: or would it be better to just use the SF page <http://sourceforge.net/projects/nazghul/>?

mistydemeo: This is what `nazgul.sf.net` redirects to, so this works, thanks!

dunn: Sorry, so you mean the archived homepage works, or the SF project page?

mistydemeo:
Looks like the test is failing.

nazghul.rb
mistydemeo: Archived homepage works!

dunn:
Yeah, I forgot to update the test after renaming stuff.

PR #254, opened 2015-05-05
puzzles 10274

dunn:
I guess the most recent Debian release is from before they switched to git, so I used that revision as the version number. The git repo has no tags yet.

The automatic .app bundle caveats made these seem redundant.


mistydemeo:
Thanks!

PR #255, opened 2015-05-08
tty-solitaire 0.1.0

iggyvolz:


tty-solitaire.rb
```diff
@@ -0,0 +1,16 @@
+class TtySolitaire < Formula
+  homepage "https://github.com/mpereira/tty-solitaire"
+  url "https://github.com/mpereira/tty-solitaire/archive/v0.1.0.tar.gz"
+  sha256 "a5d11201c163abd0e2f8c19f02a8fe4a104f1b48333fb8a267341904b2056a8d"
+
+  depends_on "ncurses"
```
mistydemeo: Can't build with the system version?

iggyvolz: It seemed to fail when I tried it without - it had an error something to the effect of "could not find libncursesx" or something of the sort - I can't remember now, I can try it again in the morning.

mistydemeo: It looks like the problem is that it specifies the ncurses library name as `ncursesw`. OS X's ncurses includes wide character support, but is installed under the name `ncurses` instead.

Setting the LDFLAGS manually worked for me:
```ruby
system "make", "LDFLAGS=-lncurses"
```

mistydemeo:
The PREFIX needs to get specified in `make install` instead of `make`; otherwise it's not being used.

I also noticed a bug where the executable is installed under the wrong name if the target directory doesn't exist; that seems to have been fixed by changes to the makefile since this release. I changed the install method to this and it works:

```ruby
  def install
    bin.mkpath
    system "make", "ldflags=-lncurses"
    system "make", "install", "PREFIX=#{prefix}"
  end
```

mistydemeo:
I submitted a fix for the ncurses issue (mpereira/tty-solitaire#8), but working around it by passing `LDFLAGS` is fine for now.

mistydemeo:
There's an 0.2.0 that fixes these issues now! No need to work around them.

iggyvolz:
@mistydemeo Awesome, I'll update the PR.

tty-solitaire.rb
```diff
@@ -0,0 +1,15 @@
+class TtySolitaire < Formula
+  homepage "https://github.com/mpereira/tty-solitaire"
+  url "https://github.com/mpereira/tty-solitaire/archive/v0.2.0.tar.gz"
+  sha256 "9d47c5a88b3d70d19acfdff29532896b7905e9ddc20b29730e3a8cb92ff3cca4"
+
+  def install
+    bin.mkpath
```
mistydemeo: Should be no longer necessary, but it's a minor quibble.

mistydemeo:
Pulled with a couple tweaks, thanks!

PR #262, opened 2015-05-27
rom-tools 0.162

mbcoguno:
`desc` DSL isn't ready yet.

### DIFF
--- a/rom-tools.rb
+++ b/rom-tools.rb
@@ -1,8 +1,8 @@
 class RomTools < Formula
   homepage "http://mamedev.org/"
-  url "https://github.com/mamedev/mame/archive/mame0161.tar.gz"
-  sha256 "f7db934676e90d0d7f2b678ccf32e580417c754dd33117ec683560956c2130b9"
-  version "0.161"
+  url "https://github.com/mamedev/mame/archive/mame0162.tar.gz"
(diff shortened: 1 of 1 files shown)


mbcoguno:
As MAME's homepage says:

> MESS is now integral part of MAME and will be distributed as such.
> As of 0.162 the MAME and MESS projects have been combined into a single emulator.

Although mame and mess still has its own source subfolder (and manfile), the build target is combined into single one, that is, they have the same execution filename `mame`. We need to decide how `mess` and

mbcoguno:
As MAME's homepage says:

> MESS is now integral part of MAME and will be distributed as such.
> As of 0.162 the MAME and MESS projects have been combined into a single emulator.

Although `mame` and `mess` still has its own source subfolder (and manfile), the build target is combined into single one, that is, they have the same execution filename `mame`. We need to decide how `mess` and `ume` formulae should be do.

1. Add `conflict_with` DSL each other.
2. Rename `mame64` to `mame`/`mess`/`ume` in their formulae with previous subtarget.
3. Build `mame` with default target (like previous `ume`) and simply delete `mess` and `ume` formulae.
4. Build `mame` with default target (like previous `ume`) and boneyard `mess` and `ume`.

BTW, I prefer 3.

Any suggestion?

mistydemeo:
3 is my preference as well.

mistydemeo:
Thanks very much, merged!

PR #264, opened 2015-05-29
Remove mess and ume formulae

mbcoguno:
The MAME and MESS projects have been combined into a single emulator.


mbcoguno:
Refer <https://github.com/Homebrew/homebrew-games/pull/262#issuecomment-105973797>

Should I open a pull request in core about `TAP_MIGRATION`?

mistydemeo:
> Should I open a pull request in core about TAP_MIGRATION?

No need to worry about that, I think. We don't need to boneyard them since the projects have been folded into mame entirely; there's no need to keep them as "broken" formulae.

PR #266, opened 2015-06-07
hedgewars formula

kodabb:


hedgewars.rb
```diff
@@ -0,0 +1,66 @@
+class Hedgewars < Formula
+  homepage "http://www.hedgewars.org"
+  desc "Hedgewars is a turn based strategy, artillery, action and comedy game"
```
mistydemeo: By convention, this goes above homepage.

hedgewars.rb
```diff
@@ -0,0 +1,66 @@
...
+    sha256 "590a9458d2123c5550d5eb39edfed28d0663703d76e13c8987237c92ca41f3f8"
+
+    patch do
+      url "http://hedgewars.org/files/patches/all.php"
+      sha256 "ee8f4e702197b7908e7fd8124fbbb54a4c7cf3792b1de2ff29a4e709ce721cbd"
+    end
+  end
+
+  head do
+    url "https://code.google.com/p/hedgewars/", :using => :hg
```
mistydemeo: This can be a single line since there's nothing else in the `head` block:

```ruby
head "https://code.google.com/p/hedgewars/", :using => :hg
```

I wonder, should we use the Github mirror, given that Google code is going away?

hedgewars.rb
```diff
@@ -0,0 +1,66 @@
...
+  depends_on "sdl_ttf"
+  depends_on "ffmpeg" if build.with? "videorec"
+  if build.with? "server"
+    depends_on "ghc" => :build
+    depends_on "cabal-install" => :build
+  end
+
+  def install
+    # rely on homebrew update, packaging, and source system
+    args = ["-DNOAUTOUPDATE=1", "-DSKIPBUNDLE=1", "-DPHYSFS_SYSTEM=1", "-DLUA_SYSTEM=1"]
```
mistydemeo: Nitpick, but this can be a `%w[]` array.

```ruby
args = %w[-DNOAUTOUPDATE=1 -DSKIPBUNDLE=1 -DPHYSFS_SYSTEM=1 -DLUA_SYSTEM=1]
```

hedgewars.rb
```diff
@@ -0,0 +1,66 @@
...
+    # homebrew fpc does not set this path in /etc/fpc.cfg
+    fpcversion = `#{"printf $(fpc -iW)"}`
```
mistydemeo: This can be simplified to:

```ruby
fpcversion = `fpc -iW`.chomp
```

hedgewars.rb
```diff
@@ -0,0 +1,66 @@
...
+    args << "-DCMAKE_Pascal_FLAGS=-Fu/usr/local/lib/fpc/" + fpcversion + "/units/x86_64-darwin/*/"
```
mistydemeo: Can't hardcode `/usr/local` - instead, can do:

```ruby
args << "-DCMAKE_Pascal_FLAGS=-Fu#{Formula["fpc"].opt_lib}/fpc/#{fpcversion}/units/x86_64-darwin/*/"
```

hedgewars.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    args << "-DGL2=1" if build.with? "gl2"
+    args << "-DNOSERVER=1" if build.without? "server"
+    args << "-DNOVIDEOREC=1" if build.without? "videorec"
+    args << "-DCMAKE_BUILD_TYPE=Debug" if build.head?
+
+    # this will prepare all server dependencies
+    if build.with? "server"
+      system "cabal", "sandbox", "init"
```
mistydemeo: Take a look at the `cabal_sandbox` mixin methods - they may help you here. Take a look at pandoc for an example.

hedgewars.rb
```diff
@@ -0,0 +1,66 @@
...
+  homepage "http://www.hedgewars.org"
+  desc "A turn based strategy/artillery/action and comedy game"
+  version "0.9.21"
+
+  stable do
+    url "http://download.gna.org/hedgewars/hedgewars-src-0.9.21.1.tar.bz2"
+    sha256 "590a9458d2123c5550d5eb39edfed28d0663703d76e13c8987237c92ca41f3f8"
+
+    patch do
+      url "http://hedgewars.org/files/patches/all.php"
```
mistydemeo: Hm, is there a versioned copy of this patch? I worry it'd change with time. The sha256 will catch it, but it'd be nice to have something stable.

kodabb: this is just a script that takes the 5 patches I have in http://hedgewars.org/files/patches and combines them in a single one. I do not plan to add any after publishing, and all patches are going away in the next version anyway

mistydemeo: OK, thanks.

hedgewars.rb
kodabb: won't that add a hard requirement in the form of
`require "language/haskell"`
?

mistydemeo: `language/haskell` is just a mixin - it provides some helpers. Using it won't require the user to have cabal/GHC installed.

hedgewars.rb
kodabb: well it's not going to go away until next year and I hope the team decides where to move in the meantime :P

kodabb:
"All is well" or so the bot reports. Anything I should do before it is merged?

mistydemeo:
✨ Thanks so much for the submission and your quick response!

kodabb:
thanks for a quick and nice review too :)

PR #268, opened 2015-06-13
Gtk+ goes quartz only

tschoonj:
This branch continues my work in Homebrew/homebrew#39868, Homebrew/homebrew-science#2311 and Homebrew/homebrew-x11#83 regarding removing X11 support in Gtk+ and its dependencies cairo and pango in favor of quartz.

I found the following packages in this tap to have a mandatory or recommended dependency on gtk+:

- [x] fceux
- [x] freeciv
- [x] pioneers 

Since these three packages are now all depending on `gtk+3` instead of `gtk+`, they can be merged in independently of the aforementioned PRs.


tschoonj:
fceux patch filed as: https://sourceforge.net/p/fceultra/bugs/703/

PR #276, opened 2015-06-25
mame 0.163

tomyun:
Following monthly mame release,

* Formula updated with `brew audit --strict`
* Patch added for clang build error: mamedev/mame@bc23c001b5dff4b271fc480c55d56785e9e3fa10


mame.rb
```diff
@@ -1,11 +1,19 @@
...
+  url "https://github.com/mamedev/mame/archive/mame0163.tar.gz"
+  version "0.163"
+  sha256 "6be1536f0c8470764aead9a32f8b5d9fd7979040b402ec337cb47f73deaa68b1"
 
   head "https://github.com/mamedev/mame.git"
 
+  # FIXME: only for clang build error on 0.163
+  stable do
+    patch do
+      url "https://github.com/mamedev/mame/commit/bc23c001b5dff4b271fc480c55d56785e9e3fa10.diff"
```
bfontaine: Please add a `sha256` for this.

tomyun: `sha256` added for the patch.

mistydemeo:
Merged with minor tweaks, thanks!

Just for future reference, when adding `stable` branches, we tend to prefer combining the `url` and `sha256` with the patch in a single `stable do` block.

PR #278, opened 2015-07-03
freeciv: Restore the GTK+2 client as the default

dtrebbien:
Depends on Homebrew/homebrew#41339 being merged.

See #268, #274


mistydemeo:
Thank you!

PR #292, opened 2015-08-11
rename: omega -> omega-rpg

xu-cheng:


mistydemeo:
How about "omega-rpg"? That's what [Debian uses](https://packages.debian.org/sid/omega-rpg).

xu-cheng:
Updated with your advice.

PR #295, opened 2015-08-14
omega-rpg: fix license location

### LINKED ISSUE #294: New omega bottle doesn't start; missing license.txt
```
lvh@zygalski ~> brew install homebrew/games/omega                                                                                                                         18:03:59
==> Installing omega-rpg from homebrew/homebrew-games
==> Downloading https://homebrew.bintray.com/bottles-games/omega-rpg-0.80.2.yosemite.bottle.tar.gz
######################################################################## 100.0%
==> Pouring omega-rpg-0.80.2.yosemite.bottle.tar.gz
🍺  /usr/local/Cellar/omega-rpg/0.80.2: 50 files, 744K
lvh@zygalski ~>                                                                                                                                                           18:04:47
lvh@zygalski ~> omega                                                                                                                                                     18:04:49

Error! File not accessible: /usr/local/Cellar/omega-rpg/0.80.2/libexec/license.txt
Further execution is impossible. Sorry.
OMEGALIB may be badly #defined in odefs.h
```

dunn:
Fixes #294.


omega-rpg.rb
```diff
@@ -29,4 +30,9 @@ def install
     bin.install "omega"
     libexec.install Dir["omegalib/*"]
   end
+
+  def post_install
```
DomT4: You should be able to do this at the end of `def install`, I'd imagine.

```ruby
libexec.install_symlink "license.txt"
```
May work. Might need the `prefix` bit before `license.txt`.

dunn: I tried that, as well as the current version, in the install block and no luck.  I assumed the code moving readmes and licenses to `prefix` was getting called after.

dunn: https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/build.rb#L136

DomT4: Argh. Yeah, it might be being [called](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/build.rb#L127-L136) as the very last thing in `install`. That's a touch annoying.

dunn: What I could do instead would be to move `license.txt` to prefix manually, then `libexec.install Dir["omegalib/*"]`, then `libexec.install_symlink "license.txt"`—that would avoid a post-install block.

DomT4: I guess if it's just the one formula acting up there's not a huge amount of point in dancing with the core code.

DomT4: > that would avoid a post-install block.

Aye. That should work, theoretically. It's up to you. I don't think it matters a huge amount if it's just the one formula; I can't imagine the `postinstall` block is going to slow the installation down by more than a few fractions of a second.

dunn: aaaa, that doesn't work either. `install_metafiles` then tries to move the symlink from `libexec` to `prefix`, and throws an error since there's already a file with the same name there (the original).

DomT4: The new `postinstall` way is failing, or the other way you'd suggested?

dunn: The way I suggested of doing it within `install`.

PR #298, opened 2015-08-18
ballerburg 1.2.0 (new formula)

tomyun:
A [classical](https://en.wikipedia.org/wiki/Ballerburg) castle combat game.


tomyun:
I couldn't come up with a nice `test` block by the way. Seems no support for any command line arguments. Any ideas?

mistydemeo:
That's perfectly fine, we can skip a test block in that case.

mistydemeo:
Thanks!

PR #299, opened 2015-08-18
liblcf 0.3 (new formula)

tomyun:
[liblcf](https://github.com/EasyRPG/liblcf) is a library required to build EasyRPG Player.


mistydemeo:
Is it feasible to write a test for this, or would it be hard to find open-licensed test data?

tomyun:
Thanks to EasyRPG/liblcf#147, the formula is updated with some tests.

- `test` block with `ToUnixTimestamp()`.
- `make check` call added.

By the way, `pkg-config --cflags liblcf` does not run correctly due to keg-only `icu4c` and `expat` dependency. Any suggestion on dealing with such a case?

fdelapena:
On our documented setup we use [PKG_CONFIG_PATH to see icu4c and expat](https://easy-rpg.org/wiki/development/compiling/liblcf/osx/autotools), but I don't know if exists a better way or convention with Homebrew.

mistydemeo:
@BrewTestBot test this please

mistydemeo:
Thanks very much!

Issue #171, opened 2014-11-09
Error installing dosbox


puffpio:
Attempting to install dosbox gives make errors. Ensured brew update and brew doctor were performed without issues

Console showed:

$ brew install dosbox
==> Installing dosbox from homebrew/homebrew-games
==> Downloading https://downloads.sourceforge.net/project/dosbox/dosbox/0.74/dos
Already downloaded: /Library/Caches/Homebrew/dosbox-0.74.tar.gz
==> ./configure --prefix=/usr/local/Cellar/dosbox/0.74 --mandir=/usr/local/Cella
==> make
make[4]: *** Waiting for unfinished jobs....
make[3]: *** [all-recursive] Error 1
make[2]: *** [all-recursive] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2

ainop:
Hi! Seems like I got into clang crash too: https://gist.github.com/anonymous/08e29cec9da480b74aae

mistydemeo:
Macports carries a couple of clang-related patches, but they're not fixing the segfaults:

* https://trac.macports.org/browser/trunk/dports/emulators/dosbox/files/src__cpu__core_dynrec__risc_x64.h.diff
* https://trac.macports.org/browser/trunk/dports/emulators/dosbox/files/patch-src-fpu-fpu_instructions_x86.h.diff

mistydemeo:
I've seen a few indications that this is a known issue when using the dynamic core in 64-bit mode, but a 32-bit build doesn't help.

mistydemeo:
However, using the `normal` core does work for me - albeit more slowly than it should. This is true with clang and gcc-4.9.

imlyc:
Hi, I got the same crash here,
    
    https://gist.github.com/dd8eeb686f64ee5ce654

Thanks.

FabianFrank:
https://gist.github.com/anonymous/318680cc3913528faba2

neoascetic:
Same issue here

mistydemeo:
Likely addressed via #310.

PR #310, opened 2015-08-21
dosbox: fix runtime crash

tomyun:
In regards with #171, the current formula builds fine, but crashes when it gets into protected mode due to dynamic core handling issue on 64-bit platform. A workaround is to set `core=normal` in the config, but gives you non-optimal performance.

I think a slow but working binary is still better than nothing, so here comes an updated formula to build one without needing manual configuration.

- `stable` builds with `--disable-dynrec` by default to prevent crash
- Add `head` build, but with no optimization due to [clang](http://sourceforge.net/p/dosbox/code-0/3894/)
- `sdl_sound` should be with `vorbis` and `ogg` as indicated [here](http://www.dosbox.com/wiki/BuildingDOSBox)
- Support `debugger` option to show a debugging console
- Add `caveats` to warn performance issue
- Add `test` block

Although it's not built optimally, many programs seemed to run okay.


mistydemeo:
@BrewTestBot test this please

mistydemeo:
Thanks very much for looking into this!

PR #311, opened 2015-08-21
dosbox-x 0.801 (new formula)

tomyun:
While playing with `dosbox` formula, I just came across with a fork called `dosbox-x` that is actively being developed recently. At least, it didn't crash with no modification of config / build options.

- `conflicts_with` `dosbox` due to same executable names
- It has additional dependency on `fluid-synth` for better MIDI support
- `test` block couldn't be put in due to a timeout kind of error


dunn:
Failing on Mountain Lion:
```
clang++ -DHAVE_CONFIG_H -I. -I../..  -I../../include -I/usr/local/include/SDL -D_GNU_SOURCE=1 -D_THREAD_SAFE  -g -O2 -std=gnu++11 -Wall -mmmx -msse -msse2 -D_FILE_OFFSET_BITS=64 -Wno-strict-aliasing  -c -o fpu.o fpu.cpp
In file included from fpu.cpp:45:
./fpu_instructions.h:387:7: error: use of undeclared identifier 'isinf'
                if (isinf(fpu.regs[st].d) && isinf(fpu.regs[other].d)) {
                    ^
./fpu_instructions.h:387:32: error: use of undeclared identifier 'isinf'
                if (isinf(fpu.regs[st].d) && isinf(fpu.regs[other].d)) {
                                             ^
2 errors generated.
```

tomyun:
Added `needs :cxx11` and `ENV.cxx11` to see if it works.

PR #314, opened 2015-08-22
supermodel 0.2a (new formula)

tomyun:
[Supermodel](http://www.supermodel3.com/) is an emulator for Sega Model 3 arcade machines.

- `Makefile` updated to grab brewed `SDL` instead of Framework.
- Also added missing labels for auto-generated code during make.
- The original program expects runtime directories (Config, Save, NVRAM) in the working directory. It's modified to use `/usr/local/var/supermodel/(Config|Save|NVRAM)`.


tomyun:
Same as #319 on the style auditing issue with build bots: ```Use %r around regular expression.```

mistydemeo:
@BrewTestBot test this please

mistydemeo:
Thanks!

re: regexes, see 576a2c2984ceb2b909cfb0993131a47cae99c9b2 - the `%r[]` regex syntax allows `/` to be used within a regex without being escaped.

Not sure why this isn't being brought up by audits running on Ruby 1.8.

tomyun:
I think it has something to do with `rubocop` updates as mentioned in Homebrew/homebrew#39620.

PR #319, opened 2015-08-23
fmsx 4.2 (new formula)

tomyun:
[fMSX](http://fms.komkon.org/fMSX/) is another MSX emulator.

Though it has `pulseaudio` dependency, sound isn't working due to its use of `/dev/dsp`, which is not available in OS X.


tomyun:
Strange style auditing issue only on Mavericks and Yosemite build bots:
```Use %r around regular expression.```

I had no issue with my local machine. I think I saw something could be related with outdated `rubocop` on CI machines that needed to be updated manually.

mistydemeo:
@BrewTestBot test this please

PR #325, opened 2015-08-24
fuse-emulator-1.1.1 (new formula)

tomyun:
[FUSE](http://fuse-emulator.sourceforge.net/) is a ZX Spectrum emulator.

- `libspectrum` is required (#324).
- Named `fuse-emulator`, not just `fuse` to avoid possible conflict.


mistydemeo:
@BrewTestBot test this please

PR #326, opened 2015-08-25
crawl 0.16.2 (new formula)

tomyun:
[Dungeon Crawl Stone Soup](https://crawl.develz.org/) is a rougelike game.

- Default built to console mode
- `with-tiles` gives nice graphics
- Sound support is implied with tiles (no way to enable only sound, no graphics)


tomyun:
Just found there is already `stone-soup`! I'll update on that formula.

mistydemeo:
Closing in favour of #327 - thanks!

PR #327, opened 2015-08-25
stone-soup 0.16.2

tomyun:
#326 replayed:

- Use brewed or OS X native deps instead of internal builds
- `with-tiles` added to support graphics mode
- Sound support is implied with tiles (no way to enable only sound, no graphics)
- `DATADIR` changed from `data` to `share` (following brewed `pkgshare`)
- Remove no longer needed SDK path
- `ENV.libstdcxx` changed to `ENV.cxx11` (only Mountain Lion needs it)
- `test` added to check version


tomyun:
Build bots failed with Jenkins error and not recovered by pushing another commits. Can someone handle this issue?

mistydemeo:
@BrewTestBot test this please

stone-soup.rb
```diff
@@ -11,25 +10,49 @@ class StoneSoup < Formula
...
 
     cd "source" do
-      # The makefile has trouble locating the developer tools for
-      # CLT-only systems, so we set these manually. Reported upstream:
-      # https://crawl.develz.org/mantis/view.php?id=7625
-      #
-      # On 10.9, stone-soup will try to use xcrun and fail due to an empty
-      # DEVELOPER_DIR
-      devdir = MacOS::Xcode.prefix.to_s
-      devdir += '/' if MacOS.version >= :mavericks && !MacOS::Xcode.installed?
```
mistydemeo: Do you mind leaving this part in? The new version still fails to build on a CLT-only system for me.

tomyun: Didn't know this issue on a CLT-only system. The formula is now updated to keep this part in. Could you restart the build bots? They failed again with a forced push commit.

mistydemeo:
Thanks!

PR #328, opened 2015-08-26
mednafen-0.9.38.5

yisraeldov:
Fixes some issues like PSX netplay and save states.

### DIFF
--- a/mednafen.rb
+++ b/mednafen.rb
@@ -2,8 +2,8 @@ require 'formula'
 
 class Mednafen < Formula
   homepage 'http://mednafen.sourceforge.net/'
-  url 'https://downloads.sourceforge.net/project/mednafen/Mednafen/0.9.36.5/mednafen-0.9.36.5.tar.bz2'
-  sha1 '02d441e18083daa539f0193121ac760882be9d19'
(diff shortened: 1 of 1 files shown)


mednafen.rb
```diff
@@ -2,8 +2,8 @@
 
 class Mednafen < Formula
   homepage 'http://mednafen.sourceforge.net/'
-  url 'https://downloads.sourceforge.net/project/mednafen/Mednafen/0.9.36.5/mednafen-0.9.36.5.tar.bz2'
-  sha1 '02d441e18083daa539f0193121ac760882be9d19'
+  url 'https://downloads.sourceforge.net/project/mednafen/Mednafen/0.9.38.5/mednafen-0.9.38.5.tar.bz2'
+  sha1 'b6dff2ff97673e4183c5628d49b6ac05c341b325'
```
mistydemeo: Mind changing this to an sha256? The hash is `10c5e6ba0d822cf2e5aaa236ca86e6ec489a56043530ba3fdc20f70929e738e8`.

yisraeldov: It's actually not working for me now, reject this and I'll make a new pull request when I get it to work right.

It needs to use GCC and not clang or get some test errors.

mistydemeo: OK, thanks!

PR #331, opened 2015-08-27
CKAN v1.14.0

hunner:
Released 4 days ago, after v1.12.0


mistydemeo:
Looks like the test is writing to the home directory, which is disallowed by the sandbox; do you think this can be worked around?

xu-cheng:
The sandbox violation is from a bug in mono. See https://github.com/Homebrew/homebrew/pull/43094

xu-cheng:

PR #332, opened 2015-08-28
mame 0.165

tomyun:
- Version bump.
- Changed to use brewed deps instead of building internally, except:
  - `expat`: brewed version not picked up automatically, os x version is old.
  - `lua`: `lua53` is not available in the core.
  - `portaudio`: though declared in `makefile`, seems not being used yet.
- `test` block improved.
- Some redundancy cleaned up.


mistydemeo:
Thanks!

mbcoguno:
Sorry, @tomyun , I have question.
Is `depends_on :python => :build` necessary? All Python code are in 3rdparty directory and seems it's not libraries that can link other formulae, it should be use system's Python.
And one more thing, due to non-universal libraries in Homebrew, `PTR64` flag is requirement for Snow Leopard. MAME use `uname` to check architecture and will cause Homebrew build failed in 10.6.

tomyun:
@mbcoguno Thanks for reporting this issue. Just opened #340 for it.

PR #335, opened 2015-08-30
gearsystem 2.1 (new formula)

tomyun:
[Gearsystem](https://github.com/drhelius/Gearsystem) is a Sega Master System / Game Gear emulator.

- Made a patch for some rendering issues (reported upstream: drhelius/Gearsystem#7)
- `CONFIG+=c++11` added to explicitly link with `libc++`, otherwise `libstdc++` will be picked.
- No `test` block, no support for command line arguments...


mistydemeo:
>  CONFIG+=c++11 added to explicitly link with libc++, otherwise libstdc++ will be picked.

This should probably only be the case if the compiler is clang on OS X 10.9+, right? Otherwise qt5 won't have been built with libc++.

mistydemeo:
Ping

tomyun:
@mistydemeo You are right. `CONFIG+=c++11` seems to pick up `libc++` for recent OSes, but not sure how it would work on 10.8 and below. Hopefully it picks up `libstdc++` by default. I just wanted to get rid of mismatching standard library warnings.

mistydemeo:
Just checked, and qt5 on 10.8 is linked against libc++. This should be fine, thanks!

PR #338, opened 2015-08-30
ufoai 2.5.0 (new formula)

tomyun:
[UFO:AI](http://ufoai.org/) is a squad-based tactical strategy game influenced by X-COM series. Since their official website lacks a stable binary package for OS X, it'd be a good addition for the tap.

One issue might be its distribution size. As maps/models data are needed in the package, its size goes almost over 1 GB.


ufoai.rb
```diff
@@ -0,0 +1,58 @@
+class Ufoai < Formula
+  desc "UFO: Alien Invasion"
+  homepage "http://ufoai.org/"
+  url "git://git.code.sf.net/p/ufoai/code", :branch => "ufoai_2.5", :revision => "1f4775cedb0937b78bf824c5c1d79efe68818ea9"
```
mistydemeo: Is there a reason you're using this instead of a 2.5.x tarball?

tomyun: Yes, the source tarball is missing some files/directories like `base` and `contrib` that seems to be hardwired with make. For example, default files for `.app` bundle are stored in `contrib/installer`. I couldn't find a better way to handle them.

mistydemeo: Ah, okay, that makes sense. I noticed there are a few tagged releases though, maybe we could use a tag? Was there a reason you picked this revision?

tomyun: I think I was a bit confused with sf.net web interface. It's now referring to the latest revision (not one revision before that I don't know why it was) using `:tag`.

mistydemeo: I think this is a branch, not a tag - I agree the UI's pretty confusing! http://sourceforge.net/p/ufoai/code/ref/master/tags/ lists a `ufoai_2.5.0` tag at a different commit than the one you switched to. The branch list [is here](http://sourceforge.net/p/ufoai/code/ref/master/branches/), for comparison.

tomyun: Sorry, I forgot to push new commit.

mistydemeo: Sorry, I mean the tag is `ufoai_2.5.0` (which is currently at `58580c9b55a69ba65fad565c7a7bd29617736f6c`); `ufoai_2.5` is a branch.

tomyun: Oh, now I remember that `ufoai_2.5` tag (`58580c`), which is a bit outdated than `ufoai_2.5.0` branch, fails to build with this formula. That's why it's instead pointing to the latest revision (`3e28f7`) of 2.5 branch, which was updated last February.

mistydemeo: Aha! Okay. I'll add a comment about that when it's merged, so we remember to move it back to a tag on the next release.

ufoai.rb
```diff
@@ -0,0 +1,57 @@
+class Ufoai < Formula
+  desc "UFO: Alien Invasion"
+  homepage "http://ufoai.org/"
+  url "git://git.code.sf.net/p/ufoai/code", :branch => "ufoai_2.5", :revision => "3e28f7cbf9f5e1cfd0fa7fdc852f833e498757c1"
+  version "2.5.0"
```
mistydemeo: Thinking about this again, do you mind changing this to reflect the fact it's not the 2.5.0 release? e.g., say `2.5.0-3e28f7cbf9f5e1`. I was going to do it myself, but it needs to get run through the bots again for them to generate new bottles reflecting the new version.

tomyun: I like your idea. Probably shorter `3e28f7` would be fine as sf.net uses.

By the way, wouldn't it be a problem if they have another updates on 2.5.0 branch, the hash happens to be something like `2xxxxx` and we want to update the formula? Of course, I hope they release 2.5.1 for that.

mistydemeo: How about including the date before the revision? Say, `2.5.0+2015-02-16-3e28f7`?

tomyun: Sounds great. Good to know another convention with `+`. Do you think it'd be enough with `2.5.0+2015-02-16` then? Since there'll be no more release from this past date.

mistydemeo: Sure, that's fine!

tomyun: Good, it's updated.

mistydemeo:
Thanks!

The large binary package may require some patience to download.

PR #341, opened 2015-08-30
rom-tools 0.165

tomyun:
Updated along with new `mame` release (#332).

- Use external deps
- With a patch preventing build failure (mamedev/mame#292)
- Added some commands for `test` block, but not much useful yet


rom-tools.rb
```diff
@@ -14,27 +14,51 @@ class RomTools < Formula
...
+    inreplace "scripts/src/main.lua", /(targetsuffix) "\w+"/, '\1 ""'
+    system "make", "TARGET=ldplayer", "TOOLS=1",
+                   "PTR64=#{MacOS.prefer_64_bit? ? 1 : 0}", # for old Macs
+                   "MACOSX_USE_LIBSDL=1",
+                   "USE_SYSTEM_LIB_ZLIB=1",
+                   "USE_SYSTEM_LIB_FLAC=1",
+                   "USE_SYSTEM_LIB_PORTMIDI=1"
     bin.install %W[
-      castool chdman floptool imgtool jedutil ldresample ldverify nltool pngcmp
+      castool chdman floptool imgtool jedutil ldplayer ldresample ldverify nltool pngcmp
```
mbcoguno: MAME have new tool named `nlwav`.

tomyun: You're right. Now it's added, but somehow it gives segmentation fault on running, like its sibling `nltool`. Any idea if it's upstream issue since they're labeled "WIP"?

PR #342, opened 2015-09-01
supertux: add 0.3.5a devel

tomyun:
- Deps cleaned up for both stable and devel
- `homepage` and `url`s now pointing to github
- `--with-apple-opengl-framework` added for stable to support OpenGL mode
- `install` cleaned up
- `test` improved
- `fails_with :clang` removed as seemingly no longer relevant (`cxx11` added in case)


supertux.rb
```diff
@@ -1,71 +1,52 @@
...
+    if build.stable?
+      # https://trac.macports.org/ticket/29635
+      inreplace "src/menu.h", /Menu::(get_controlfield_key_into_input)/, '\1'
+      system "./configure", "--disable-dependency-tracking",
+                            "--with-apple-opengl-framework",
                             "--prefix=#{prefix}"
-      system "make install"
+      system "make", "install"
+    else
+      ENV.cxx11
```
mistydemeo: I think this is unrelated to the clang issue; I'm able to build using clang without it. I think it may have been a different bug that was fixed in this release.

tomyun: Okay, I just added it to make sure it builds on older versions of OS X, but maybe we want to give it a shot without one? `cmake` seems to check CXX11 support:
```
-- Performing Test COMPILER_SUPPORTS_CXX11
-- Performing Test COMPILER_SUPPORTS_CXX11 - Success
```

mistydemeo:
@BrewTestBot test this please

tobbi:
Would it be possible for you to include this PR https://github.com/SuperTux/supertux/pull/82 ? I believe, this was merged post-0.3.5a

Direct link to patch: https://patch-diff.githubusercontent.com/raw/SuperTux/supertux/pull/82.patch

tomyun:
@tobbi added the patch, but it still seems not working on Mavericks.

tomyun:
@tobbi I made a small patch to remove non-const copy constructor and assign operator for `Action` class, and now it builds on Mavericks:
```
inreplace "src/supertux/screen_manager.hpp", /.*\(Action &a\)[^}]+}/, ""
```

Seems they conflict with const copy constructors in some cases, but have no idea why they have no problem with other versions of OS. Could you review this modification upstream?

tobbi:
Sorry, I am totally unfamiliar with this code. Are we talking about these here:
https://github.com/SuperTux/supertux/commit/1c67f68d48a605e1cb5fd6c6c888219565f2bcc1#diff-5ae278e3fa72bb8cc401258d22f943d1R86

If so, I will put them inside #ifdef's

tomyun:
The regex snippet above is to remove `Action(Action &a) : ... { ... }` and `Action& operator=(Action &a) { ... }` in the code you link. Seems these non-const operators prevented implicit definition of const operators and thus caused some const violation, leading to build failure. I have no idea if they are really meant to do something in the code, while the comment just says that added for MSVS compatibility, so hopefully fairly safe?

tobbi:
I hope the above commit fixes the issue. I have set those two copy constructors inside #ifdef statements.

tobbi:
Until we release 0.4.0, please go with the manual replacement set in the formula.

dunn:
@tobbi, have your suggested changes been made in this PR?

tobbi:
@dunn Well, my suggested changes are on top of master right now, not in the 0.3.5a release. If you want, you can try to pull in the patch from here and apply it, and then see whether compilation works:
https://github.com/SuperTux/supertux/commit/d8f76e2e1af3bda02a3c6d3b03d6ec371740bb5b.diff

But, the inreplace suggestion might work as well.

bfontaine:
Ping?

tobbi:
See if it works with inreplace, please.

bfontaine:
The patch doesn’t work; let’s keep the `inreplace` until the next release.

tobbi:
Sure, sounds fine with me.

bfontaine:
@BrewTestBot test this please.

bfontaine:
Thank you @tomyun for your contribution to Homebrew; we appreciate it! :tada:

PR #343, opened 2015-09-02
vecx 1.0 (new formula)

tomyun:
[vecx](https://github.com/jhawthorn/vecx) is a [Vectrex](https://en.wikipedia.org/wiki/Vectrex) emulator.

- No `test` block.


vecx.rb
```diff
@@ -0,0 +1,17 @@
...
+  url "https://github.com/jhawthorn/vecx/archive/v1.0.tar.gz"
+  sha256 "6aac0c1840982aa3cc8607d5fa65597b6644840f2ba1ea3715a0349522312f56"
+  head "https://github.com/jhawthorn/vecx.git"
+
+  depends_on "sdl"
+  depends_on "sdl_gfx"
+  depends_on "sdl_image" if build.head?
+
+  def install
+    inreplace ["e6809.c", "vecx.c"], "__inline", "static inline"
```
mistydemeo: Do you mind reporting this issue to the developers?

vecx.rb
```diff
@@ -0,0 +1,17 @@
+class Vecx < Formula
+  desc "Vectrex emulator"
+  homepage "https://github.com/jhawthorn/vecx"
+  url "https://github.com/jhawthorn/vecx/archive/v1.0.tar.gz"
+  sha256 "6aac0c1840982aa3cc8607d5fa65597b6644840f2ba1ea3715a0349522312f56"
+  head "https://github.com/jhawthorn/vecx.git"
+
+  depends_on "sdl"
+  depends_on "sdl_gfx"
+  depends_on "sdl_image" if build.head?
```
mistydemeo: Can you scope this in a `head` block instead?

```ruby
head do
  url "https://github.com/jhawthorn/vecx.git"
  depends_on "sdl_image"
end
```

vecx.rb
tomyun: Reported to jhawthorn/vecx#3. But `diff` couldn't be usable with `patch` here since `master` branch has different whitespace than the current version.

vecx.rb
tomyun: Done!

mistydemeo:
Thanks!

PR #347, opened 2015-09-04
nxengine 1.0.0.6 (new formula)

tomyun:
[NXEngine](http://nxengine.sourceforge.net/) is a rewrite of freeware [Cave Story](http://www.cavestory.org/) game, originally known as [Dokutsu Monogatari](http://studiopixel.sakura.ne.jp/archives/index.html). English patched [resource](http://agtp.romhack.net/project.php?id=cavestory) is generally required to run the game.

The formula automatically download resource and place it in `var/nxengine` for further extraction. The original engine assumes everything in the current working directory, so doesn't fit nicely with brewed layout, so a bit of patch has been made to handle locations.

Other notable things:
- Added [replacement](https://gist.github.com/yinyin/2027912) of endian.h for OS X.
- Removed unused Linux header that led to build failure.
- Symlink system font, instead of copying Courier New bundled with the archive, avoiding any possible license issue.
- No `test` block, no support for any command line arguments.


mistydemeo:
@BrewTestBot test this please

PR #348, opened 2015-09-05
mgba 0.3.0 (new formula)

tomyun:
[mGBA](http://mgba.io/) is a Game Boy Advance emulator.

We can build two different frontends with SDL and Qt. Qt frontend has more features and still has a dependency on SDL. When it's selected (by default), the CLI binary of SDL frontend (`mgba`) will be replaced by a script for running Qt app (`mGBA`).


dunn:
QT5 has been having a lot of issues lately; will try building on El Capitan again after 5.5.1 is release: https://github.com/Homebrew/homebrew/issues/44150#issuecomment-141722083

PR #352, opened 2015-09-05
c2048 0+20150805 (new formula)

tomyun:
[2048.c](https://github.com/mevdschee/2048.c) is a console version of the game [2048](http://gabrielecirulli.github.io/2048/), which is again a clone of [Threes](http://asherv.com/threes/).

- Named `c2048` instead of unacceptable `2048.c`.
- Version `0+20150805` made up with the date of latest revision. No official release so far and looks like it's not going to happen.


dunn:
From the README:
```
wget https://raw.githubusercontent.com/mevdschee/2048.c/master/2048.c
gcc -o 2048 2048.c
./2048
```
Let's follow that and name this 2048.

tomyun:
I don't think Homebrew, or Ruby, supports class name starting with numbers.

dunn:
Oh, I didn't think of that. What about '2048c'? That's closer to the original.

tomyun:
'2048c' still starts with numbers, so I couldn't make it working.

dunn:
Right, still not thinking straight, sorry.

mistydemeo:
@BrewTestBot test this please

PR #354, opened 2015-09-05
angolmois 2.0alpha2 (new formula)

tomyun:
[Angolmois](http://mearie.org/projects/angolmois/) is a rhythm game that plays [BMS](https://en.wikipedia.org/wiki/Be-Music_Source) songs.


mistydemeo:
@BrewTestBot test this please

PR #355, opened 2015-09-06
gnu-go: add HEAD, fix readline

tomyun:
- add `HEAD`
- fix `--with-readline` argument to correctly link against `libedit`
- add `test`


dunn:
Merged in 896912fddb67c31027a3303b1897121c0cde2705, thanks so much.

PR #356, opened 2015-09-06
gnu-chess: add HEAD

tomyun:
- add `HEAD`.
- remove unrecognized argument `--disable-debug`.
- remove `doc.install` as they are not real docs, but source files for making docs.
- modify `test` to better handle versions, e.g, comparing 6.2 vs 6.2.1 would have still passed.


dunn:
@BrewTestBot test this please

dunn:
Merged in d302380ed0abb0280948721d4e659f05ef74db5a ♔♕

PR #358, opened 2015-09-06
openttd 1.5.2

tomyun:
- version bump
- add `openttd` script to run app bundle
- update `test` block to check version


dunn:
@BrewTestBot test this please

dunn:
Failing to build on El Capitan:
```
/private/tmp/openttd20151003-52479-18vhtww/openttd-1.5.2/src/music/cocoa_m.cpp:76:4: error: no matching function for call to 'AUGraphNodeInfo'
                        AUGraphNodeInfo(graph, node, &desc, &unit);
                        ^~~~~~~~~~~~~~~
```

tomyun:
They used `__AUDIOCOMPONENT_H__` macro to determine if the header is from 10.6 and above, but El Capitan seemed to rename it to `AudioUnit_AudioComponent_h`. A simple patch could bypass this issue, but still stuck with another error with undeclared `CMGetSystemProfile` and `CMCloseProfile`. Maybe better wait for upstream fixes.

bfontaine:
You could bump the `opengfx` resource to 0.5.3 too.

tomyun:
Obsolete by #460.

PR #359, opened 2015-09-06
open-tyrian: add HEAD, update URLs

tomyun:
- URLs changed from Google code to Bitbucket
- `head` added
- `test` added
- data directory handling improved

By the way, feels like it should have been named `opentyrian`, not `open-tyrian`.


dunn:
@BrewTestBot test this please

dunn:
Merged in c3e30654a873546efa582588002908a101d94291, thanks 💖

We can do formula renames now, but I'm not sure if this is important enough to warrant one.  You could add an alias, though; we support those in Taps now: https://github.com/Homebrew/homebrew/pull/43851

PR #360, opened 2015-09-06
xrick 021212 (new formula)

tomyun:
[xrick](http://www.bigorno.net/xrick/) is an open source clone of [Rick Dangerous](https://en.wikipedia.org/wiki/Rick_Dangerous).


mistydemeo:
@BrewTestBot test this please

PR #361, opened 2015-09-06
fceux: patch SConstruct to point to correct LUA5.1 include folder

keymone:
builds successfully and fceux menu now displays Load LUA script (i haven't actually tried running anything yet).

### DIFF
--- a/fceux.rb
+++ b/fceux.rb
@@ -19,6 +19,7 @@ class Fceux < Formula
   depends_on "sdl"
   depends_on "libzip"
   depends_on "gtk+3" => :recommended
+  depends_on "lua51" => :build
 
   # Make scons honor PKG_CONFIG_PATH and PKG_CONFIG_LIBDIR
   # Reported upstream: https://sourceforge.net/p/fceultra/bugs/625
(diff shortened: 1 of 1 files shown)


fceux.rb
```diff
@@ -74,19 +73,15 @@ def install
...
-+  BoolVariable('LUA',       'Enable Lua support', 0),
-   BoolVariable('GTK', 'Enable GTK2 GUI (SDL only)', 1),
-   BoolVariable('GTK3', 'Enable GTK3 GUI (SDL only)', 0),
-   BoolVariable('NEWPPU',    'Enable new PPU core', 1),
+@@ -140,7 +140,7 @@ else:
+     lua_available = False
+     if conf.CheckLib('lua5.1'):
+       env.Append(LINKFLAGS = ["-ldl", "-llua5.1"])
+-      env.Append(CCFLAGS = ["-I/usr/include/lua5.1"])
++      env.Append(CCFLAGS = ["-I/usr/local/include/lua5.1"])
```
dunn: Has this been/can this be reported upstream?

mistydemeo: We can't hardcode `/usr/local` - users may have Homebrew installed in other locations.

keymone: thanks for feedback, i've updated the patch to use HOMEBREW_PREFIX.

i have no idea if any of this can/will be pushed upstream, i just found this project today.

PR #363, opened 2015-09-07
qtads 2.1.6 (new formula)

tomyun:
[QTads](http://qtads.sourceforge.net/) is a multimedia interpreter for TADS games built on Qt GUI.


mistydemeo:
@BrewTestBot test this please

PR #374, opened 2015-09-09
sz81 2.1.7 (new formula)

tomyun:
[sz81](http://sz81.sourceforge.net/) is a Sinclair [ZX80](https://en.wikipedia.org/wiki/ZX80)/[ZX81](https://en.wikipedia.org/wiki/ZX81) emulator.


mistydemeo:
@BrewTestBot test this please

PR #375, opened 2015-09-10
opensyobon 1.0rc2 (new formula)

tomyun:
[Open Syobon Action](http://sourceforge.net/projects/opensyobon/) is an open-source clone of [Syobon Action](https://en.wikipedia.org/wiki/Syobon_Action) which is a parody of Super Mario Bros.

- The actual versioning was just `rc2` released after `rc1` and `0.9`, so here I just assumed it's going to be a release candidate for `1.0`. I don't think we're going to see any new release anyways. Just for history keeping stuff.


mistydemeo:
@BrewTestBot test this please

PR #376, opened 2015-09-11
qdae 0.0.10 (new formula)

tomyun:
[QDAE](http://www.seasip.info/Unix/QDAE/) is an emulator for [Apricot computers](https://en.wikipedia.org/wiki/Apricot_Computers).

- It supports a limited set of command line arguments, but couldn't find a clever way to incorporate them into `test`.


mistydemeo:
@BrewTestBot test this please

PR #380, opened 2015-09-12
hatari 1.9.0

tomyun:
[Hatari](http://hatari.tuxfamily.org/) is an emulator for Atari ST/STE/TT/Falcon.

`hatari` package is currently served in [homebrew-x11](https://github.com/Homebrew/homebrew-x11) tap, but it doesn't seem to have x11 dependency, at least for OS X app. I'm going to open a PR that migrates it from `homebrew-x11` to `homebrew-games`.

Here is a set of changes in the formula:
- Version bump (1.8.0 -> 1.9.0)
- Modernize formula.
- Install .app bundle which has much richer UI.
- Install `tos.img` directly from [EmuTOS](http://emutos.sourceforge.net/), not using bundled one. `HEAD` install breaks because it doesn't have one in the repo. Other than that, they are identical binaries.
- Drop `:x11` dependency.
- Add dependency on `portaudio` used for microphone input.
- Remove unnecessary cmake args. It seems to build fine without them.
- Add `test` block.


dunn:
@BrewTestBot test this please

hatari.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def install
+    # Set .app bundle destination
+    inreplace "src/CMakeLists.txt", "/Applications", prefix
+    system "cmake", *std_cmake_args
+    system "make"
+    prefix.install "src/Hatari.app"
+    bin.write_exec_script "#{prefix}/Hatari.app/Contents/MacOS/hatari"
+    resource("emutos").stage do
```
dunn: Since this is only necessary for HEAD, we can wrap this block in `if build.head?`

tomyun: Actually, copying tos.img into app bundle is still needed for `stable`; `head` needs to download itself beforehand. I thought it'd be okay for both to use the same `resource`, otherwise we'd need two `install` statements with conditional `build.stable?`.

mistydemeo:
@BrewTestBot test this please

mistydemeo:
Thanks!

PR #382, opened 2015-09-13
magnetix 3.1 (new formula)

tomyun:
[magnetiX](http://www.maczentrisch.de/magnetiX/) is an interpreter for the games from [Magnetic Scrolls](https://en.wikipedia.org/wiki/Magnetic_Scrolls).


mistydemeo:
Thanks!

PR #383, opened 2015-09-14
pytouhou 634 (new formula)

tomyun:
[PyTouhou](http://pytouhou.linkmauve.fr/) is an engine for playing [Touhou 6](http://en.touhouwiki.net/wiki/Embodiment_of_Scarlet_Devil) shoot'em up game.

- `version` follows revision number of hg repo as [the author](http://hg.linkmauve.fr/touhou/diff/26d9d251f658/setup.py) and [other distro]
(https://aur.archlinux.org/packages/pytouhou-hg/) do.
- Explicit dep on `lha`: it doesn't seem to get automatically installed when extracting `.lzh` resource.
- `inreplace`
  - `setup.py` can't determine hg revision number as `.hg` directory is not available when staging
  - Override default game path to `pkgshare/game` where demo is installed at.


tomyun:
```Error: /usr/local/opt/py3cairo not present or broken```

I encountered this problem on my local machine too. `pycairo3` should have been automatically pulled when `pygobject3` gets installed with `--with-python3`. Is this known homebrew issue?

dunn:
Yes, I've run into similar issues.

mistydemeo:
@BrewTestBot test this please

mistydemeo:
I fixed `lha` not being inferred as a dependency in Homebrew.

tomyun:
Then can I remove an explicit dep on `lha` from the formula? Also curious if the `pycairo3` issue mentioned above is still relevant.

mistydemeo:
> Then can I remove an explicit dep on lha from the formula?

Yes, you can!

The `pycairo` issue is still relevant. It's an issue with the way optional dependencies are expressed.

Jack has covered it in more detail, but this is similar to the issue in Homebrew/homebrew#20276: the class is first loaded with the dependencies not specified, because they are a language-level conditional and `build.include?` for the option will evaluate false at that point in time. This causes problems later in the build process when the class is instantiated.

mistydemeo:
Thanks!

PR #385, opened 2015-09-15
nesemu2 0.6.1+20140930 (new formula)

tomyun:
[nesemu2](https://github.com/holodnak/nesemu2) is an NES emulator.

- The repo has no release tag, but its website and [source code](https://github.com/holodnak/nesemu2/blob/master/source/version.h) indicates that the last release was 0.6.1. Since then its development was sporadic, thus `version` becomes `0.6.1+20140930` including some patches.
- `head` has new commits this week, but fails to build yet.


mistydemeo:
@BrewTestBot test this please

PR #389, opened 2015-09-18
liblcf 0.3.2

tomyun:

PR #390, opened 2015-09-18
easyrpg-player 0.3.2

tomyun:
Probably should work without #389 this time, but would recommend to bottle it after `liblcf` gets update.


dunn:
Merged in 9c0b0caaa3895f5967f70a007393e2841098eac4 👾

PR #391, opened 2015-09-18
angband 4.0.2

tomyun:
- version bump
- now `autogen` required for `stable`
- a bit cleanup on `x11` option and `args`
- add `--with-sdl` option for SDL frontend, though it's not very usable (too laggy)
- add `--with-cocoa` option to build native Cocoa app bundle


dunn:
Merged in 680410cb58a5b572dcc6972d7b2bf3cba67b2d7d, thanks again!

PR #393, opened 2015-09-19
mupen64plus: merge mupen64plus-video-glide64mk2

tomyun:
`mupen64plus-video-glide64mk2` is actually already provided with `mupen64plus` source bundle since version 2.0.0, so it's redundant to have another formula, which is outdated anyways.

- `mupen64plus-video-glide64mk2`
  - formula removed
  - (automatic) version bump (2.0 to 2.5)
- `inreplace` added to correctly set `-mmacosx-version-min`, otherwise it'll link against libstdc++ on recent OSes, giving warnings on dependencies with libc++.
- `homepage` updated, `desc` added


mupen64plus.rb
```diff
@@ -31,6 +33,9 @@ class Mupen64plus < Formula
   end
 
   def install
+    # Prevent different C++ standard library warning
+    inreplace Dir["source/mupen64plus-**/projects/unix/Makefile"], /(-mmacosx-version-min)=\d+\.\d+/, "\\1=#{MacOS.version}"
```
dunn: If this can't be set with a make flag, it should probably be reported upstream. Thanks!

mistydemeo:
Mind rebasing this?

tomyun:
Rebased onto the master.

mistydemeo:
Thanks!

PR #404, opened 2015-09-21
gnuski 0.3 (new formula)

tomyun:
[GNUSki](http://sourceforge.net/projects/gnuski/) is an open source clone of [SkiFree](https://en.wikipedia.org/wiki/SkiFree) that runs on the terminal.


gnuski.rb
```diff
@@ -0,0 +1,15 @@
...
+  url "https://downloads.sourceforge.net/project/gnuski/gnuski/gnuski-0.3/gnuski-0.3.tar.gz"
+  sha256 "1b629bd29dd6ad362b56055ccdb4c7ad462ff39d7a0deb915753c2096f5f959d"
+
+  def install
+    system "make"
+    bin.install "gnuski"
+  end
+
+  test do
+    File.executable? "#{bin}/gnuski"
```
bfontaine: Please write a test for the actual functionality.

gnuski.rb
```diff
@@ -0,0 +1,15 @@
+class Gnuski < Formula
+  desc "Open source clone of Skifree"
+  homepage "http://gnuski.sourceforge.net/"
+  url "https://downloads.sourceforge.net/project/gnuski/gnuski/gnuski-0.3/gnuski-0.3.tar.gz"
+  sha256 "1b629bd29dd6ad362b56055ccdb4c7ad462ff39d7a0deb915753c2096f5f959d"
+
+  def install
+    system "make"
+    bin.install "gnuski"
```
bfontaine: Is there a `make install` we can use?

tomyun: No, there is no `install` target.

gnuski.rb
tomyun: It's a very simple program that accepts no arguments. I'll remove `test` block then.

gnuski.rb
```diff
@@ -0,0 +1,11 @@
...
+  end
```
bfontaine: Please add a `test do` block.

tomyun: I have no idea for writing a `test do` for programs that accept no arguments. Any suggestion?

mistydemeo:
If the program is interactive-only, I think it's fine to have no test. Unfortunately common in this tap!

PR #406, opened 2015-09-23
moon-buggy: add devel, head

tomyun:
- add `devel` and `head`
- `test` updated
- some formula modernization

also tried to support sound add-on, but it crashed on runtime.


dunn:
Merged in 48a76526b2161ce9cdd294150bc4e0c0979af856! 🚀

PR #408, opened 2015-09-25
blastem 0.3.1 (new formula)

tomyun:
[BlastEm](http://rhope.retrodev.com/files/blastem.html) is an emulator for Sega Genesis games.

- no source archive available, so directly pulled from repo
- missing tag in the repo, so `:revision` picked up manually by looking up comments
- requires `shaders` in the current directory, so a wrapper script is written to do chdir.


dunn:
The build timed out at 1 hour. How long does it take for you to compile locally?

tomyun:
Should be built within minutes; took me just seconds on my machine. The culprit was `test` popping a modal dialog to display help message. It somehow seems to have different behaviors depending how it's called. `test` is now updated to ensure `headless` flag is set.

blastem.rb
```diff
@@ -0,0 +1,24 @@
+class Blastem < Formula
+  desc "Fast and accurate Genesis emulator"
+  homepage "http://rhope.retrodev.com/files/blastem.html"
+  url "http://rhope.retrodev.com/repos/blastem", :using => :hg, :revision => "c9ed929ee984"
```
bfontaine: Doesn’t it have tagged releases?

mistydemeo:
Ping.

blastem.rb
mistydemeo: I checked into it, looks like upstream didn't tag the 0.3.1 release (but the commit message makes it clear that is 0.3.1). Will add a comment to that effect.

mistydemeo:
@BrewTestBot test this please

mistydemeo:
Looks like the install is missing the `rom.db`, so this won't run any games as-is.

tomyun:
- updated to install `rom.db`
- script modified to run on the current directory

The default `rom.db` is not mandatory and contains only a small list of games by the way. Probably the culprit was `bin/blastem` that changed the current directory to `pkgshare`, making relative paths unworkable.

blastem.rb
```diff
@@ -0,0 +1,24 @@
...
+  version "0.3.1"
+  head "http://rhope.retrodev.com/repos/blastem"
+
+  depends_on "pkg-config" => :build
+  depends_on "sdl2"
+  depends_on "glew"
+
+  def install
+    system "make"
+    pkgshare.install %w[blastem default.cfg rom.db shaders]
```
dunn: Can `blastem` be in `libexec` instead?

tomyun: That would make more sense. Updated.

mistydemeo:
Thanks!

Issue #394, opened 2015-09-20
mame incomplete installation?


CRP:
Hi, currently the mame formula installs the binary and man page. Should it also install latest versions of the content of folders such as hash, hlsl, samples, keymaps etc.? These are needed for the correct functioning of mame, but currently need to be downloaded/updated manually.

dunn:
We'd be happy to take a look at a pull request!  See https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md and let us know if you have questions!

mistydemeo:
As of #413, these extra files are installed in `share/mame`.

PR #413, opened 2015-09-30
mame 0.166

tomyun:
- monthly version bump
- install extra stuff (#394)
  - except `hlsl` (Windows-only)
  - instead use `shader` (glsl)


mbcoguno:
Can you add `depends_on :python => :build if MacOS.version <= :snow_leopard` for older Macs that ship Python 2.6, not 2.7?

tomyun:
Sure, it's now added.

dunn:
And merged in 19a6c02e1368880dab3665c75d380bd009bf9b3a, thanks!

PR #421, opened 2015-10-06
ppsspp 1.1.0

tomyun:
- version bump
- add new deps: `glew`, `snappy`
- use `DUSE_SYSTEM_FFMPEG=ON` instead of `inreplace`
- add `test` to bypass audit warning


ppsspp.rb
```diff
@@ -34,4 +30,8 @@ def install
       bin.write_exec_script "#{prefix}/PPSSPPSDL.app/Contents/MacOS/PPSSPPSDL"
     end
   end
+
+  test do
+    File.executable? "#{prefix}/PPSSPPSDL.app/Contents/MacOS/PPSSPPSDL"
```
bfontaine: Please add a test for the actual functionality.

tomyun: I'd love to make one, but Mac version of ppsspp doesn't support any command line arguments, so couldn't come up with a proper one. Do you think we'd better remove this block then?

bfontaine: Yeah, it’s better to remove it.

tomyun: It's now removed.

dunn:
Merged in 5f97986c2f4e936a1ef701f0db474e9415710f3c, thanks again!

PR #422, opened 2015-10-06
fceux: remove unused libzip

tomyun:
Refer to Homebrew/homebrew#44656.

Initially thought `revision` update for this formula, but it seems `fceux` has no actual dependency on `libzip`. Maybe confused with `minizip` which is currently built internally by default?


bfontaine:
Thank you for your contribution @tomyun; we appreciate it! :smile:

PR #429, opened 2015-10-20
dwarf-fortress: freetype caveat

### LINKED ISSUE #418: Add building `freetype` with `--universal` to `dwarf-fortress` caveats
See Homebrew/homebrew#44570

timvisher:
Closes #418

/cc @bfontaine

### DIFF
--- a/dwarf-fortress.rb
+++ b/dwarf-fortress.rb
@@ -17,8 +17,12 @@ class DwarfFortress < Formula
   end
 
   def caveats; <<-EOS.undent
-      If you're using a retina display, the PRINT_MODE should
-      be changed to STANDARD in #{libexec}/data/init/init.txt
+    If you're using a retina display, the PRINT_MODE should
+    be changed to STANDARD in #{libexec}/data/init/init.txt
+
+    Dwarf Fortress currently references freetype at a deprecated
+    location if installed on 10.11. `brew reinstall freetype
+    --universal` may help.
     EOS
   end
 end


timvisher:
@bfontaine Let me know why the build is failing and I can try to fix it.

bfontaine:
@timvisher Do you mind replacing the `sha1` with a `sha256`? It fails because of that :/

timvisher:
@bfontaine Attempted. We'll see how that goes.

dunn:
Should there be something like 
```ruby
depends_on "freetype" => [:optional, :universal]
```

timvisher:
@dunn Sure? Let me know and I'll add it.

dunn:
What does dwarf-fortress use freetype for?

timvisher:
@dunn I'm assuming for rendering? I actually have no idea.

dunn:
We should figure out if it's a mandatory or optional dependency before doing anything, then.

timvisher:
@dunn It's mandatory afaict. Homebrew/homebrew#44570

dunn:
Oh, cool. Then go ahead and add `depends_on "freetype" => :universal` and remove the caveat. Thanks!

dwarf-fortress.rb
```diff
@@ -18,6 +18,10 @@ def install
   def caveats; <<-EOS.undent
     If you're using a retina display, the PRINT_MODE should
     be changed to STANDARD in #{libexec}/data/init/init.txt
+
+    Dwarf Fortress currently references freetype at a deprecated
```
mistydemeo: If it's compatible with a Homebrew-installed freetype, how about we make the dep mandatory and rewrite the install names in the binary to point it at that?

mistydemeo:
Alternately, if the issue is that it needs X11 and the X11 path changed, maybe we can rewrite them to point to the actually-installed X11's prefix instead.

dwarf-fortress.rb
timvisher: > If it's compatible with a Homebrew-installed freetype, how about we make the dep mandatory and rewrite the install names in the binary to point it at that?

I have no idea how to rewrite the binary to point to that. :P

timvisher:
@mistydemeo @dunn The recipe did have a dep declared on X11, so I would think it doesn 'need X11'. Pretty much all I know about this is that I had a broken dwarf-fortress game which started working as soon as I reinstalled `freetype` with the `--universal` flag. I'm happy to either continue to have recommendations thrown my way or to have someone who knows more than I do take this over.

timvisher:
ping :)

bfontaine:
Sorry I modified the formula; could you please rebase the PR on top of `master`?

timvisher:
@bfontaine Done.

mistydemeo:
After a little experimentation, I have an alternate suggestion in #444; this uses `install_name_tool` to point SDL_ttf at the right library regardless of X11 prefix.

PR #443, opened 2015-11-07
hedgewars: Update stable version to 0.9.22

kodabb:
The patches are not needed any more as they all have been applied.

### DIFF
--- a/hedgewars.rb
+++ b/hedgewars.rb
@@ -5,16 +5,10 @@ class Hedgewars < Formula
 
   desc "A turn based strategy/artillery/action and comedy game"
   homepage "http://www.hedgewars.org"
-  version "0.9.21"
 
   stable do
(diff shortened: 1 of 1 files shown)


hedgewars.rb
```diff
@@ -5,16 +5,11 @@ class Hedgewars < Formula
 
   desc "A turn based strategy/artillery/action and comedy game"
   homepage "http://www.hedgewars.org"
-  version "0.9.21"
+  version "0.9.22"
```
dunn: We don't need to specify the version explicitly anymore; Homebrew can figure this one out from the URL. Thanks!

kodabb:
updated

dunn:
Thanks! Merged in a06564000d696877622df447621520298b54c090.

PR #447, opened 2015-11-13
supertux 0.3.5a (devel)

bfontaine:
This aims to replace #342.


bfontaine:
The patch fails to apply.

PR #456, opened 2015-12-02
dwarf-fortress 0.42.1

bfontaine:
It builds but crash at launch. Does anyone knows how to get more info on the crash?

```
$ dwarffortress
Loading bindings from data/init/interface.txt
2015-12-02 23:28:05.400 dwarfort.exe[63642:6486113] 23:28:05.400 WARNING:  140: This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h.
2015-12-02 23:28:06.045 dwarfort.exe[63642:6486113] Warning once: This application, or a library it uses, is using NSQuickDrawView, which has been deprecated. Apps should cease use of QuickDraw and move to Quartz.
New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

Resetting textures
```

It opens a window then closes it immediatly and returns. I was able to get an Apple crash report at some point but I can’t anymore for some reason.


mistydemeo:
Are you on a retina display? The default "2D" display mode crashes on higher-DPI devices. I get the same behaviour if I don't make the config change suggested by the caveats.

bfontaine:
Nope I’m on a mid '13 MacBook Air.

mistydemeo:
Hm, interesting! Does changing the display mode not fix it?

bfontaine:
@mistydemeo Yes it does; should we `inreplace` it by default then?

mistydemeo:
@bfontaine </late> Probably a good idea, yeah. Skipped doing it before because it only seemed to be causing issues on retina devices then.

bfontaine:
Thanks! I removed the caveats and added an `inreplace` call.

BlueChili:
Root of the problem comes from df packaging an old version of libsdl, attempting to use NSQuickDrawView, fixed following the recommendation by lethosor at http://www.bay12games.com/dwarves/mantisbt/view.php?id=2624

mistydemeo:
@bfontaine Want to rebase this on the latest master then merge?

bfontaine:
@mistydemeo Sorry for the delay; #459 updated it to 0.42.02 so this PR is now irrelevant.

PR #467, opened 2015-12-20
supertux 0.4.0

maxteufel:
- `homepage` now pointing to our domain
- support for building 0.1.3 removed, 0.4.0 is new stable
- removed `inreplace ...`, which is said to be not needed anymore (however should be tested since I don't have a Mavericks machine)

### DIFF
--- a/supertux.rb
+++ b/supertux.rb
@@ -1,67 +1,37 @@
 class Supertux < Formula
   desc "Classic 2D jump'n run sidescroller game"
-  homepage "https://supertux.github.io/"
+  homepage "http://supertuxproject.org/"
+  url "https://github.com/SuperTux/supertux/releases/download/v0.4.0/supertux-0.4.0.tar.bz2"
+  sha256 "d18dde3c415e619b4bb035e694ffc384be16576250c9df16929d9ec38daff782"
 
-  stable do
-    url "https://github.com/SuperTux/supertux/releases/download/v0.1.3/supertux-0.1.3.tar.bz2"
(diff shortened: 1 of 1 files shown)


supertux.rb
```diff
@@ -1,25 +1,10 @@
...
-  end
-
-
-  devel do
-    url "https://github.com/SuperTux/supertux/releases/download/v0.3.5a/supertux-0.3.5a.tar.bz2"
-    sha256 "71c34502b5879a4130fb111ac754967c3a632ef3017f30883a4d496d87bef8c7"
+    url "https://github.com/SuperTux/supertux/releases/download/v0.4.0/supertux-0.4.0.tar.bz2"
+    sha256 "d18dde3c415e619b4bb035e694ffc384be16576250c9df16929d9ec38daff782"
 
     depends_on "cmake" => :build
```
dunn: If these are needed for any build type now, you can drop the `stable do` entirely and just have:
```ruby
[...]
  homepage "http://supertuxproject.org/"
  url "https://github.com/SuperTux/supertux/releases/download/v0.4.0/supertux-0.4.0.tar.bz2"
  sha256 "d18dde3c415e619b4bb035e694ffc384be16576250c9df16929d9ec38daff782"

  depends_on "cmake" => :build
[...]

maxteufel: Fixed.

dunn:
Thanks so much! Merged in 1ded6b840281c9ec311a2b2ab9676c35e9c57b49.

PR #366, opened 2015-09-07
fizmo 0.7.9 (new formula)

tomyun:
[Fizmo](https://christoph-ender.de/fizmo/) is a [Z-Machine](https://en.wikipedia.org/wiki/Z-machine) interpreter.

- ~~requires `homebrew/dupes/ncurses` due to its dependency on `ncursesw`~~
- use system `ncurses` as `ncursesw` (Homebrew/homebrew#32947, Homebrew/homebrew#38355)

PR #477, opened 2015-12-30
fizmo 0.7.10 (new formula)

tomyun:
Updated from #366.

- version bump
- `sdl2` instead of `sdl`

Not sure if the `test` issue still persists with the bot.


mistydemeo:
Hm, looks like the test still fails.

mistydemeo:
It's also working for me when I test locally.

mistydemeo:
Without being able to reproduce it myself, I'm going to guess the issue is happening here:

```c
  // newterm instead of initscr to init ncurses without clearing screen
  // so we can exit gracefully in case of errors.
  if ((s = newterm(NULL, stdin, stdout)) == NULL)
    return -1;
```

e.g., `newterm()` returns `NULL`, presumably because the way in which the program is being run by the test bot renders it impossible to create a curses session.

tomyun:
Probably since bots are being run headless? Now I'm thinking of keeping `test` for only console version then. Do you think it's acceptable?

mistydemeo:
That's fine, yes! Thank you.

tomyun:
Updated.

mistydemeo:
Thanks!

PR #478, opened 2015-12-30
fs-uae 2.6.2 / 2.7.6dev

tomyun:
- version bump for both `stable` and `devel`
- `libmpeg2` added (no longer built by itself)


mistydemeo:
Looks like it can't find openal on 10.9:

```
libfsemu/src/emu/audio_openal.c:18:10: fatal error: 'al.h' file not found
#include <al.h>
         ^
1 error generated.
```

mistydemeo:
It works with openal-soft on Mavericks; maybe add a conditional `depends_on "openal-soft" if MacOS.version <= :mavericks`?

mistydemeo:
Thanks!

PR #357, opened 2015-09-06
pc6001vx 2.1 (new formula)

tomyun:
[PC6001VX](http://eighttails.seesaa.net/) is an emulator for NEC [PC-6001](https://en.wikipedia.org/wiki/NEC_PC-6001) computer.

PR #479, opened 2015-12-30
pc6001vx 2.11 (new formula)

tomyun:
Updated from #357:
- `head` added
- `patch` removed (merged upstream)
- `test` removed (no use)


mistydemeo:
Thanks!

PR #494, opened 2016-01-28
rom-tools 0.170

tomyun:
- version bump
- `test` update


dunn:
@BrewTestBot test this please

PR #501, opened 2016-02-12
jpcsp 0.7+20160203 (new formula)

tomyun:
[jpcsp](http://jpcsp.org/) is a PSP emulator written in Java.


dunn:
Could we get a test? Thanks!

tomyun:
I wanted too, but couldn't come up with a good one given that it's a Java-based GUI app that doesn't receive any meaningful command-line arguments.

dunn:
Fair enough.  Merged in 921083548385e6c59273de989122c8fa4927f281, thanks again!

PR #502, opened 2016-02-13
raine 0.64.10 (new formula)

tomyun:
[Raine](http://raine.1emulation.com/) is an arcade emulator.

- `patch` included (zelurker/raine#5)
- all `universal` deps since raine builds only on `m32`


dunn:
```
==> audit problems
homebrew/games/raine:
 * GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)
```
I'll let @mistydemeo decide if we should waive the notability requirement here.

tomyun:
I believe Raine has been around since 90s and notable enough to many arcade lovers. Their GitHub repo looks like new because it's set up quite recently, I guess.

mistydemeo:
Haha, yes, I remember using it in high school! They seem to have mirated from a self-hosted git repo to Github recently.

PR #503, opened 2016-02-15
ppsspp 1.2

tomyun:
- version bump
- `libzip` dep added with a trick (Homebrew/homebrew-nginx#49)
- rename exec script from `PPSSPPSDL` to `ppsspp`


dunn:
Merged in 3c587d550c59caaf356dee707ae145c3041538ea, thanks!

PR #506, opened 2016-02-22
Revision for ffmpeg 3.0 update

tomyun:
In regards with Homebrew/homebrew#49178, there are three formulae need rebottling: `mgba`, `pc6001vx`, and `ppsspp`. `hedgewars` has optional dep.


mistydemeo:
Thanks!

PR #507, opened 2016-02-23
DUNGEON 4.0 (new formula)

GOFAI:
The classic text adventure that served as the basis of the Zork series, updated to compile under gfortran.

The original code made extensive use of a nonstandard FORTRAN extension that conflated logical and bitwise operators, but other than correcting for this the code should be identical to that written by Robert Supnik in the 1980s.

```brew audit --strict --online``` complains about the presence of tabs in the formula, but these are part of a patch to the VERY old-fashioned FORTRAN source code that is necessary to tell the executable to look for the data files where homebrew installed them rather than the pwd.

### CODE REVIEW

File: dungeon.rb [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+  url "https://github.com/GOFAI/dungeon/archive/4.0.tar.gz"
+  sha256 "be2217be9e23861f22c14c4395da272fca5fb08a1741f52fd393792908279bea"
+
+  depends_on :fortran
+
+  def install
+    chdir "src" do
+      # look for game files where homebrew installed them, not pwd
+      inreplace "game.f" do |s|
+        s.gsub! "FILE='dindx',STATUS='OLD',", "FILE='/usr/local/opt/dungeon/share/dindx',"
```
mistydemeo: Instead of hradcoding the path, can you interpolate in the directory name? e.g.

```ruby
s.gsub! "FILE='dindx',STATUS='OLD',", "FILE='#{pkgshare}/dindx',"
```

Similar with the line below.

mistydemeo:
From the audit:

> * Use spaces instead of tabs for indentation

Mind updating the indentation in the formula?

Don't mind the notability part, I can ignore that.

mistydemeo:
Sorry, missed your comment re: the tabs. If the source contains literal tabs that makes sense.

dungeon.rb
GOFAI: I just tried this but it causes gfortran to choke because it's very sensitive to the exact formatting using these legacy settings (i.e., it won't accept lines that wouldn't fit on a punch card):
```
==> make
gfortran -g -std=legacy   -c -o dungeon.o dungeon.f
gfortran -g -std=legacy   -c -o game.o game.f
gfortran -g -std=legacy   -c -o gdt.o gdt.f
gfortran -g -std=legacy   -c -o objects.o objects.f
gfortran -g -std=legacy   -c -o parser.o parser.f
game.f:220:24:

 10000 OPEN (UNIT=1,FILE='/usr/local/Cellar/dungeon/4.0/share/dungeon/dindx',
                        1
Error: Invalid value for FILE specification at (1)
game.f:221:6:

  1 STATUS='OLD',FORM='FORMATTED',
      1
Error: Bad continuation line at (1)
game.f:221:6: Error: Unclassifiable statement at (1)
game.f:228:27:

  OPEN (UNIT=DBCH,FILE='/usr/local/Cellar/dungeon/4.0/share/dungeon/dtext',
                           1
Error: Invalid value for FILE specification at (1)
game.f:229:6:

  1 STATUS='OLD',FORM='UNFORMATTED',ACCESS='DIRECT',
      1
Error: Bad continuation line at (1)
game.f:229:6: Error: Unclassifiable statement at (1)
game.f:203:72: Error: Label 10000 referenced at (1) is never defined
make: *** [game.o] Error 1
make: *** Waiting for unfinished jobs....
```
The interpolated path runs past the line length limit and also causes the line continuation to fail. While inelegant, hardcoding the path ensures that it's a known length that won't cause this issue.

mistydemeo: Is there a way we could patch this that wouldn't run into this problem? We allow users to customize Homebrew's install location; it's not guaranteed to be at `/usr/local`.

In a pinch you could use `opt_pkgshare` instead of `share`, which would use the `/usr/local/opt` symlink instead of the full path. That could still mean some prefixes would be too long to fit on a punchcard, but it'd be an improvement.

GOFAI:
OK, I've edited the formula to use #{opt_share} instead of the hardcoded path.

mistydemeo:
Just out of curiosity, is there a reason you switched from `opt_pkgshare` to `opt_share`?

GOFAI:
It turns out using ```opt_pkgshare``` resulted in ```/usr/local/opt/dungeon/share/dungeon/dindx``` instead of ```/usr/local/opt/dungeon/share/dindx```, so while the patch compiled fine the resulting executable looked for the files in the wrong locations.

dungeon.rb
```diff
@@ -0,0 +1,35 @@
...
+        s.gsub! "FILE='dindx',STATUS='OLD',", "FILE=#{opt_share}/dindx',"
+        s.gsub! "1	FORM='FORMATTED',ACCESS='SEQUENTIAL',ERR=1900)", "1	STATUS='OLD',FORM='FORMATTED',
+	2	ACCESS='SEQUENTIAL',ERR=1900)"
+        s.gsub! "FILE='dtext',STATUS='OLD',", "FILE='#{opt_share}/dtext',"
+        s.gsub! "1	FORM='UNFORMATTED',ACCESS='DIRECT',", "1	STATUS='OLD',FORM='UNFORMATTED',ACCESS='DIRECT',"
+      end
+      system "make"
+      bin.install "dungeon"
+    end
+    share.install "dindx"
```
mistydemeo: What about installing these to `pkgshare`? Linking them to `/usr/local/share/dungeon` seems a bit clearer.

GOFAI: You mean change ```#{opt_share} to ```/usr/local/share/dungeon``` and install the data files to ```pkgshare```? I just tried that on my machine and it all seems to work correctly.

mistydemeo: Yes, please!

mistydemeo: Or, rather, change `#{opt_share}` to `#{opt_pkgshare}` and install the data files to `pkgshare`.

mistydemeo:
Thank you! Squashed to one commit and merged.

I realized there was a solution to silence the tab warning, so I fixed that up in 43da4fb08f03b5e46933a50dc343a3449947df7a.

GOFAI:
Thanks!

PR #514, opened 2016-03-03
Update Gearsystem to version 2.2

drhelius:
Updated to latest release.


dunn:
Possible to add a quick test while you're at it?

drhelius:
This is a desktop emulator, I'm afraid I don't know what to test.

dunn:
Something like `--version` is better than nothing; are there any basic non-interactive commands?

drhelius:
Nope, if you run it from the command line the application just opens and run in the foreground until the user closes it. I'm afraid there aren't any parameters.

dunn:
OK, merging as-is.  Thank you for contributing!

PR #517, opened 2016-03-06
scummvm-tools 1.8.0 (new formula)

tomyun:
A collection of [tools](http://wiki.scummvm.org/index.php/User_Manual/Appendix:_Tools) for ScummVM.

### DIFF
--- a/scummvm-tools.rb
+++ b/scummvm-tools.rb
@@ -1,0 +1,24 @@
+class ScummvmTools < Formula
+  desc "Collection of tools for ScummVM"
+  homepage "http://www.scummvm.org/"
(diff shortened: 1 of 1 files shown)


dunn:
Would it make sense for this to be included in the ScummVM formula?

tomyun:
Maybe we'd better have a separate formula since they have different source archive and dependencies.

tomyun:
By the way, `scummvm-tools` have own release schedule than main `scummvm`. It seems to be much sporadic.

dunn:
@brewtestbot test this please

dunn:
Merged as d7426298d0017519a365cb310658d166069d255d, thanks!

Issue #496, opened 2016-01-28
cataclysm HEAD fails to compile


derickvn:
When compiling the HEAD version of cataclysm, the build fails with the output below. Building without the '--HEAD' option works fine.

Output:
```
brew install homebrew/games/cataclysm --with-tiles --HEAD
==> Installing cataclysm from homebrew/games
==> Cloning https://github.com/CleverRaven/Cataclysm-DDA.git
Updating /Library/Caches/Homebrew/cataclysm--git
==> Checking out branch master
==> make NATIVE=osx RELEASE=1 OSX_MIN=10.11 CXX=clang++ LD=clang++ CXXFLAGS=  -I
Last 15 lines from /Users/derick/Library/Logs/Homebrew/cataclysm/01.make:
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DOSX_SDL2_LIBS -DTILES -DLOCALIZE -I../src -Wno-unused-variable -c src/activity_handlers.cpp -o obj/tiles/activity_handlers.o
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DOSX_SDL2_LIBS -DTILES -DLOCALIZE -I../src -Wno-unused-variable -c src/activity_item_handling.cpp -o obj/tiles/activity_item_handling.o
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DOSX_SDL2_LIBS -DTILES -DLOCALIZE -I../src -Wno-unused-variable -c src/addiction.cpp -o obj/tiles/addiction.o
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DOSX_SDL2_LIBS -DTILES -DLOCALIZE -I../src -Wno-unused-variable -c src/advanced_inv.cpp -o obj/tiles/advanced_inv.o
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DOSX_SDL2_LIBS -DTILES -DLOCALIZE -I../src -Wno-unused-variable -c src/ammo.cpp -o obj/tiles/ammo.o
#define VERSION "0.C-9789-g14d21ff"
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DOSX_SDL2_LIBS -DTILES -DLOCALIZE -I../src -Wno-unused-variable -c src/animation.cpp -o obj/tiles/animation.o
In file included from src/animation.cpp:9:
src/cata_tiles.h:4:10: fatal error: 'SDL.h' file not found
#include <SDL.h>
[...]

Issue #484, opened 2016-01-08
brew install homebrew/games/cataclysm --HEAD 


superwg1984:
brew install homebrew/games/cataclysm --HEAD
==> Installing cataclysm from homebrew/games
==> Cloning https://github.com/CleverRaven/Cataclysm-DDA.git
Updating /Library/Caches/Homebrew/cataclysm--git
==> Checking out branch master
==> make NATIVE=osx RELEASE=1 CXX=clang++ LD=clang++ CXXFLAGS=  CLANG=1
Last 15 lines from /Users/wanggeng/Library/Logs/Homebrew/cataclysm/01.make:
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DLOCALIZE  -c src/wincurse.cpp -o obj/wincurse.o
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DLOCALIZE  -c src/wish.cpp -o obj/wish.o
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DLOCALIZE  -c src/worldfactory.cpp -o obj/worldfactory.o
Cannot run an astyle check, your system either does not have astyle, or it is too old.
clang++  -DRELEASE -DMACOSX -DGIT_VERSION -DLOCALIZE  -c src/version.cpp -o obj/version.o
[...]

PR #521, opened 2016-03-19
cataclysm: remove flags overriding

tomyun:
In regards with #496, the culprit seemed to be our `CXXFLAGS` that wiped out internal parameters populated by `sdl2-config`. When I removed `CXXFLAGS`, it just built fine for both `stable` and `HEAD` with/without `tiles` option on OS X 10.11.

Removing `CXX` and `LD` didn't show any side effect either. If this could be a regression for older platforms, please let me know.

We could also get rid of `OSX_MIN` in the next release; it's upstream thanks to #484.


dunn:
@BrewTestBot test this please

dunn:
Sorry for the slow response; merged as ca44dc9e1bc73f7c83ebd135586306c910f064ae.

PR #522, opened 2016-03-19
liblcf 0.4.1

tomyun:


tomyun:
Looks like CI now runs automatic tests on other formulae that depends on it. I guess it should be fine if #523 gets merged. To merge that, however, we should merge this PR first. Chicken or the egg?

dunn:
Easiest way to get around the chicken-egg problem is to have both commits in the same pull request.

tomyun:
Good to know! I've just pushed them together. Thanks.

PR #526, opened 2016-03-21
easyrpg-player & liblcf 0.4.1

tomyun:
Pushing #522 and #523 together to avoid the chicken-egg problem.


tomyun:
@dunn looks like `easyrpg-player` got built ahead `liblcf, still failed with missing bottle?

dunn:
Try having one commit per formula, with liblcf first.

tomyun:
@dunn well, that's how I did earlier but somehow it didn't work. Although `liblcf` was committed first, the bot picked up `easyrpg-player` then failed as we saw. Later I combined them in one commit, but still with no luck. I'll split the commits anyways to see how it turns out this time.

dunn:
Oh, still the same error.  This is very strange.

tomyun:
@dunn could you manually merge them separately as we did in the previous releases?

dunn:
I can't merge anything that doesn't pass the CI, since we want bottles as well.

tomyun:
@dunn it seems bottling was not an issue with logs from #522. `liblcf` bottles were built successfully there, but only the last dependency check failed since the current `easyrpg-player` couldn't be built with the latest `liblcf` we just bottled. It will be easily resolved once we merge `liblcf` bottles first, and then queue up `easyrpg-player` bottling again, as we did for the previous releases.

dunn:
OK, looks like both bottles were created successfully, so I merged this as c3ef91a686cf573f47f7296746394b51569fed60 and 0a8e89fb052b408731009b4b4f23c900243a437e.  Let me know if you see any problems, and thanks again!

tomyun:
`easyrpg-player` seems to run without problem. Thanks!

PR #532, opened 2016-03-29
New Formula: Libggz

s172262:
It may make libggz, libggzcore and libggzmod.


s172262:
Should I separate this request  to two parts?

dunn:
They can be the same PR since one depends on the other, but two commits, thanks!

If you were closing/opening to re-queue a build, please don't.  Just ask us and we'll do it.

s172262:
Thank you for your advice. I would try in other branch.

PR #533, opened 2016-03-29
nethack: Add a recipe for a fork of 3.6.0 with GUIs

mikeage:
This fork offers graphical interfaces (both SDL2 and Qt4) for MacOS X.

### DIFF
--- a/nethack-interfaces.rb
+++ b/nethack-interfaces.rb
@@ -1,0 +1,68 @@
+class NethackInterfaces < Formula
+  desc "Single-player roguelike video game"
+  homepage "http://www.nethack.org/index.html"
+  version "3.6.0-chasonr.20160329"
+  url "https://github.com/chasonr/nethack-3.6.0-patches.git", :branch => "interfaces"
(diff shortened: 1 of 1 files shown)


mikeage:
This recipe works, but I'm pretty sure there's a lot of room for improvement in the naming. Suggestions welcome. 

Note that the official code for 3.6.0 does not support tiled interfaces on Mac OS, so there's no way to implement this in the main version.

RikerW:
@mikeage it apparently failed the checks?

mikeage:
Correct... it's nowhere near popular enough.

RikerW:
Is that what the check looks for?  And it should be added anyway IMO.

mikeage:
Yes. It'd be nice, but the official policy is pretty clear :-/

apjanke:
Yeah, this is not popular enough to live in one of the official Homebrew taps. This is a good candidate for [your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md), or maybe a community tap for NetHack variants.

PR #534, opened 2016-03-30
New Formula: libggz and ggz-client-libs

s172262:
This is retry of #532. I'm sorry to have caused so much trouble of this.


s172262:
It seems that I need to try one by one. I would try only libggz.

PR #539, opened 2016-03-31
openttd 1.6.0-RC1 (devel)

tomyun:


openttd.rb
```diff
@@ -3,7 +3,7 @@ class Openttd < Formula
   homepage "https://www.openttd.org/"
   url "https://binaries.openttd.org/releases/1.5.3/openttd-1.5.3-source.tar.xz"
   sha256 "d8b9a7aaca7c9f3ff69b1d210daf1e2658402941bb9b30cb2789a9df73d1ba63"
-  revision 1
+  revision 2
```
dunn: Can you make this (and the `write_exec_script`) a separate commit? Thanks!

tomyun: Now they're in two separate commits.

PR #543, opened 2016-04-04
New Formula: Widelands

s172262:


widelands.rb
```diff
@@ -0,0 +1,49 @@
...
+    ENV.append "LDFLAGS", "-L#{Formula["gettext"].opt_prefix}/lib -L#{Formula["lua51"].opt_prefix}/lib"
+
+    cd "build" do
+      system "cmake", "..", "-DLUA_INCLUDE_DIR:PATH=#{Formula["lua51"].opt_prefix}/include/lua-5.1",
+                      "-DLUA_LIBRARIES:STRING=#{Formula["lua51"].opt_prefix}/lib/liblua.5.1.dylib",
+                      "-DLUA_LIBRARY:FILEPATH=#{Formula["lua51"].opt_prefix}/lib/liblua.5.1.dylib",
+                      "-DINTL_INCLUDE_DIR:PATH=#{Formula["gettext"].opt_prefix}/include",
+                      # Without the follewing option, Cmake uses the library of framework.
+                      "-DPNG_PNG_INCLUDE_DIR:PATH=#{Formula["libpng"].opt_prefix}/include",
+                      "-DWL_INSTALL_DATADIR=share/games/widelands", "-DWL_INSTALL_BINDIR=bin", *std_cmake_args
```
dunn: Does `DWL_INSTALL_DATADIR` have to be a relative path?  Ideally we should use `pkgshare` here.

widelands.rb
```diff
@@ -0,0 +1,49 @@
...
+      system "make", "install"
+    end
+  end
+
+  def caveats; <<-EOS.undent
+    Data files install in #{HOMEBREW_PREFIX}/share/games/widelands.
+    When you play, use the follewing data-dir option.
+
+      $ widelands --datadir=#{HOMEBREW_PREFIX}/share/games/widelands
```
dunn: If this flag will need to be passed to every invocation of `widelands`, we can write a wrapper script for it.  Install the real binary in `libexec` instead, then do something like
```ruby
(bin/"widelands").write <<-EOS.undent
#!/bin/sh
exec #{libexec}/widelands --datadir=#{pkgshare}/data "$@"
EOS
```
See [`git-ftp.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/git-ftp.rb) and [`antlr.rb`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/antlr.rb) for more examples.

widelands.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  EOS
+  end
+
+  test do
+    system "#{HOMEBREW_PREFIX}/bin/widelands", "--version"
```
dunn: This can just be `system bin/"widelands", "--version"`.

But what would be great would be a test that checks more of the functionality.

widelands.rb
s172262: In CMakeLists.txt
> `#` Widelands data directory
> if (NOT WL_INSTALL_DATADIR OR WL_INSTALL_DATADIR STREQUAL "")
> 	set (WL_INSTALL_DATADIR share/games/widelands)
> endif (NOT WL_INSTALL_DATADIR OR WL_INSTALL_DATADIR STREQUAL "")

But this implementation is incompetent. 
If **WL_INSTALL_DATADIR STREQUAL** was not declared, data files are installed in #{prefix}/ without "data" directory.
So I set `-DWL_INSTALL_DATADIR=share/games/widelands` to follow the plan of the Widelands producer. 
But I have no policy. I would change to `pkgshare/"data"`.

widelands.rb
s172262: In the game mode Widelands needs mouse operations at the startup screen to run the main functions.
I have no idea to check whether Widelands normally works in game mode.

widelands.rb
s172262: It's so cool idea!

widelands.rb
dunn: Please do then, thanks!

dunn:
Looks like you accidentally pulled in some unrelated commits.  Could you rebase this against the latest `master`?

PR #546, opened 2016-04-08
New Formula: Widelands

s172262:
I'm sorry to break previous pull request #543.


widelands.rb
```diff
@@ -0,0 +1,44 @@
...
+  depends_on "sdl_mixer"
+  depends_on "sdl_net"
+  depends_on "sdl_ttf"
+  depends_on "sdl_gfx"
+  depends_on "doxygen"
+  depends_on "glew"
+  depends_on "lua51"
+
+  def install
+    ENV.append "CPPFLAS", "-I#{Formula["gettext"].opt_prefix}/include -I#{Formula["lua51"].opt_prefix}/include/lua-5.1"
```
dunn: Should've caught this last time, but you can use `opt_include` and `opt_lib` here and below to save space.

dunn: But are you sure these two lines are necessary? It builds for me without them.

widelands.rb
```diff
@@ -0,0 +1,44 @@
...
+    ENV.append "LDFLAGS", "-L#{Formula["gettext"].opt_prefix}/lib -L#{Formula["lua51"].opt_prefix}/lib"
+
+    cd "build" do
+      system "cmake", "..", "-DLUA_INCLUDE_DIR:PATH=#{Formula["lua51"].opt_prefix}/include/lua-5.1",
+                      "-DLUA_LIBRARIES:STRING=#{Formula["lua51"].opt_prefix}/lib/liblua.5.1.dylib",
+                      "-DLUA_LIBRARY:FILEPATH=#{Formula["lua51"].opt_prefix}/lib/liblua.5.1.dylib",
+                      "-DINTL_INCLUDE_DIR:PATH=#{Formula["gettext"].opt_prefix}/include",
```
dunn: It seems to build and run fine without this line or the `libpng` one below as well.

widelands.rb
```diff
@@ -0,0 +1,44 @@
...
+                      # Without the following option, Cmake uses the library of framework.
+                      "-DPNG_PNG_INCLUDE_DIR:PATH=#{Formula["libpng"].opt_prefix}/include",
```
s172262: In my local Mac libpng**14** of Mono framework was linked without this line.

widelands.rb
s172262: I think that it is necessary because gettext is `keg_only`.

widelands.rb
s172262: I think that it is necessary because gettext is `keg_only`.

widelands.rb
```diff
@@ -0,0 +1,44 @@
...
+  depends_on "sdl_net"
+  depends_on "sdl_ttf"
+  depends_on "sdl_gfx"
+  depends_on "doxygen"
+  depends_on "glew"
+  depends_on "lua51"
+
+  def install
+    ENV.append "CPPFLAS", "-I#{Formula["gettext"].opt_prefix}/include -I#{Formula["lua51"].opt_prefix}/include/lua-5.1"
+    ENV.append "LDFLAGS", "-L#{Formula["gettext"].opt_prefix}/lib -L#{Formula["lua51"].opt_prefix}/lib"
```
s172262: In my local Mac Cmake linkded with #{HOMEBREW_PREFIX}/lib/liblua.dylib which is symbolic link of liblua.5.3.dylib without  "-L#{Formula["lua51"].opt_prefix}/lib"
Formula of lua51 didn't have liblua.dylib.

widelands.rb
s172262: Cmake couldn't find out lua.hpp . So I think that Cmake didn't see #{Formula["lua51"].opt_include}/lua5.1 or #{Formula["lua51"].opt_include}/lua-5.1.

widelands.rb
dunn: Homebrew adds all the dependencies to the required `PATH`s, so CMake should be able to find it.

widelands.rb
dunn: Oh got it, with `lua53` installed it's picking up that instead.  But we should be able to remove the `gettext` stuff.

apjanke:
@BrewTestBot test this please

apjanke:

For future reference, you can squash your commits to a single commit (to keep history tidy) with `git rebase -i master` or `git commit --amend`.

PR #550, opened 2016-04-20
gnu-chess 6.2.2

tomyun:


dunn:
Is there really no chess emojis?

Anyway thanks, merged as 06f75e57a702323b4b1bef79ee5e59384cd051a8!

tomyun:
There are [Unicode chess symbols](https://en.wikipedia.org/wiki/Chess_symbols_in_Unicode), but no fancy emojis from Apple yet? ♖♟

dunn:
:D

PR #553, opened 2016-04-27
mame 0.173

tomyun:


dunn:
Thanks as always! Merged as 3b90f883f403b8cf71b96ddf3e8636ee075377d3.

PR #554, opened 2016-04-29
mkxp 0+20160224 (new formula)

tomyun:
[mkxp](https://github.com/Ancurio/mkxp) is an open source interpreter for games made with RPG Maker XP/VX/VX Ace.

There's no tagged release yet nor planned anytime soon (Ancurio/mkxp#30), but `mkxp` has been around for years and already distributed with a commercial game port like [To the Moon](http://store.steampowered.com/app/206440/), so I'd call it stable enough for packaging.

Dependencies are somewhat hardwired and that's another good reason to have this formula in the tap; it'd be quite a lot of work to have all deps ready and there's no latest binary release for OS X recently.

### DIFF
--- a/mkxp.rb
+++ b/mkxp.rb
@@ -1,0 +1,74 @@
+class Mkxp < Formula
+  desc "RPG Maker XP/VX/VX Ace games interpreter"
+  homepage "https://github.com/Ancurio/mkxp"
+  url "https://github.com/Ancurio/mkxp.git", :revision => "47ef36ca190a8be7449aae4717af9e70a74cbf19"
+  version "0+20160224"
+  head "https://github.com/Ancurio/mkxp.git"
+
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "cmake" => :build
+  depends_on "libtool" => :build
+  depends_on "pkg-config" => :build
+  depends_on "qt" => :build
+
+  depends_on "boost"
+  depends_on "fluid-synth" => :optional
+  depends_on "libguess"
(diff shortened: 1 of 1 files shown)


tomyun:
By the way, I couldn't come up with a good `test` case. Probably need an actual resource/script that exits automatically, but not sure how I could make/get one.

apjanke:
@BrewTestBot test this please

apjanke:
In that thread on Mac OS X building instructions, [upstream says](https://github.com/Ancurio/mkxp/issues/30#issuecomment-189685188) there are no plans for tagged releases, and stuff is still broken.

> >   By the way, does mkxp have a plan to release any official version? It'd be nice to have a tagged release with some reasonable versioning scheme. Homebrew may not accept HEAD-only packages in their games tap.
>
> Not in the near future. As long as vital functionality such as font rendering is still broken, it makes little sense to me to start maintaining releases. Besides that, I'd like to focus the little time I spent on the project on writing / fixing code instead of version bureaucracy.

I think that means this isn't going to make the cut for official Homebrew taps, and it should continue to live in your own tap until upstream is prepared to bless a labeled release.

I see you've got a formula [here](https://github.com/tomyun/homebrew-games/blob/mkxp-20160104/mkxp.rb) in your fork of the homebrew/games tap. I'd recommend creating a new tap of your own (e.g. tomyun/games-personal or tomyun/games-ty) independent of homebrew-games so it can be the root of its own fork network, and make it clear there are standalone formulae there.

Thank you for your submission, regardless!

PR #568, opened 2016-05-22
blastem 0.4.0

tomyun:
The new version provides GUI with its own rom `menu.bin`. To build this rom, [vasm](http://sun.hasenbraten.de/vasm/) and [xcftools](http://henning.makholm.net/software) are added for build-time resources.

There's an open PR for `xcftools` in the core tap: Homebrew/homebrew-core#1216.

I'm unsure if we could submit a PR for `vasm` too as it looks very customizable and each tool may require different configurations at its build-time.


tomyun:
Close in favor of #616.

PR #573, opened 2016-05-28
cdogs-sdl 0.6.1 (new formula)

tomyun:
[C-Dogs SDL](https://cxong.github.io/cdogs-sdl/) is an open source, classic overhead run-and-gun game.

No `test` block is there yet; command line arguments like `help` somehow didn't work on the shell.

### DIFF
--- a/cdogs-sdl.rb
+++ b/cdogs-sdl.rb
@@ -1,0 +1,24 @@
+class CdogsSdl < Formula
+  desc "Classic overhead run-and-gun game"
+  homepage "https://cxong.github.io/cdogs-sdl/"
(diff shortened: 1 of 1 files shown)


apjanke:
We can test that the thing launches without crashing, [the way `pioneers` does](https://github.com/Homebrew/homebrew-games/blob/f378195bcbb6a9855764a339c4413cc13eb370f0/pioneers.rb#L40-L47).

```
  test do
    server = fork do
      system "#{bin}/pioneers-server-console"
    end
    sleep 5
    Process.kill("TERM", server)
  end
```

I've done so in a new PR; we'll see how it works: #654

apjanke:
Merged in #654. Thank you for your contribution to Homebrew!

PR #579, opened 2016-06-10
instead 2.4.1 (new formula)

tomyun:
[INSTEAD](http://instead.syscall.ru/index.html) is the interpreter of simple text adventures.

### DIFF
--- a/instead.rb
+++ b/instead.rb
@@ -1,0 +1,25 @@
+class Instead < Formula
+  desc "Interpreter of simple text adventures"
+  homepage "https://instead.syscall.ru/"
+  url "https://github.com/instead-hub/instead/archive/2.4.1.tar.gz"
+  sha256 "2368d6e7a38ce1e82feea3848d10916f52caf6103579ac5fbc2bd6dcfdf89c6e"
+  head "https://github.com/instead-hub/instead.git"
+
+  depends_on "cmake" => :build
+  depends_on "lua"
+  depends_on "sdl"
+  depends_on "sdl_image"
+  depends_on "sdl_mixer"
+  depends_on "sdl_ttf"
+
+  def install
+    mkdir "build" do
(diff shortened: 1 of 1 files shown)


dunn:
@brewtestbot test this please

dunn:
@BrewTestBot test this please

dunn:
This is on the verge of notability so that's good enough for me.

instead.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def install
+    mkdir "build" do
+      system "cmake", "..", "-DWITH_GTK2=OFF", *std_cmake_args
+      system "make", "install"
+    end
+  end
+
+  def caveats
+    "If the game path is not recognized, try using a full path."
```
dunn: Not sure what this means.  Can you add a shim so this isn't necessary?

tomyun: If I remember correctly, commands like `instead .` or `instead ./game` din't work as expected, but had to be `instead /Users/user/games/game`. I'm afraid I can't come up with a shim for this case.

dunn: Is that because of how Homebrew installs it, or a general bug?

tomyun: I don't think it's related to Homebrew installation. I'd not call it a bug either because Instead primarily uses its own directory at user's home to keep game files. If games are stored there, users may not need to deal with paths in general.

dunn: If it's not Homebrew-specific, it doesn't need to be in the caveats.  But I would consider it a bug, and the fact that users need to be alerted to it seems to support that.

tomyun: When users follow the [instructions](https://github.com/instead-hub/instead), there should be no problem. I'd just remove caveat for now then.

tomyun:
Seems like it recently became *notable enough* with 50 stars. No more audit errors.

dunn:
Merged as 39f3e410ce83e7e095b969ab4569ffb1ea8ad940, thanks again!

PR #582, opened 2016-06-25
mednafen: revision for gcc 6

tomyun:
1. Basically a revision to rebottle mednafen for the new GCC 6 (Homebrew/homebrew-core#676). Otherwise, it would stop working after upgrading to GCC 6 and cleaning up GCC 5.
2. `fails_with :clang` updated to reflect the latest version (703).
3. Fix a build error with GCC 6, which was addressed [upstream](http://forum.fobby.net/index.php?t=msg&th=1318).
4. Fix audit warning on `to_s`.


dunn:
Thanks! Merged as edeb868c08551be298f7b8e80c63a78408ea744e.

PR #584, opened 2016-06-26
fmsx 4.6

tomyun:
- version bump
- ROM files are no longer bundled with the source archive, thus pulled as `resource`.


dunn:
Thanks❗ Merged as 917d51927ba1bf0f050c31f8b53373d2035f65af.

PR #594, opened 2016-07-06
dwarf-fortress 0.43.05

sandlst:
Previous versions were 32 bit only; starting with this release the default download is a 64 bit build, whereas the 32 bit build can be retrieved as df_43_05_osx32.tar.bz2.


dunn:
Thank you! Merged as 5b0aab42272677015e63e1f7f933d2ebb3057bfb.

PR #598, opened 2016-07-11
go-gui: remove weird version specifier in Java dependency

apetresc:
I'm actually not sure what this thing was even *trying* to accomplish; normally this is used for setting a scope, not a version. It was failing for me on a machine with the latest `brew cask install java` before, and now it's fine.

In any case, it's certainly *NOT* true that `go-gui` requires Java 1.6 specifically to run. Any later versions (like the latest 1.8) work just fine, and you'd be hard-pressed to find a Mac with Java <= 1.5 anyway.

### DIFF
--- a/go-gui.rb
+++ b/go-gui.rb
@@ -12,7 +12,7 @@ class GoGui < Formula
   end
 
   depends_on :ant => :build
-  depends_on :java => "1.6"
+  depends_on :java
 
   resource "quaqua" do
     url "http://www.randelshofer.ch/quaqua/files/quaqua-5.4.1.nested.zip"
--- a/go-gui.rb
+++ b/go-gui.rb
@@ -12,7 +12,7 @@ class GoGui < Formula
   end
 
   depends_on :ant => :build
-  depends_on :java
+  depends_on :java => "1.6+"
 
   resource "quaqua" do
     url "http://www.randelshofer.ch/quaqua/files/quaqua-5.4.1.nested.zip"


apetresc:
Can anyone help me figure out why the tests failed for this patch? The formula works fine for me, and `brew audit go-gui` and `brew test go-gui` both report no problems, even in a headless environment.

go-gui.rb
```diff
@@ -12,7 +12,7 @@ class GoGui < Formula
   end
 
   depends_on :ant => :build
-  depends_on :java => "1.6"
+  depends_on :java
```
dunn: With the `:java` dependency you can do `depends_on :java => "1.6+"` to indicate 1.6 or higher;.

apetresc: Ah, I understand. So the problem here isn't the version specifier, it's that the specifier is set to `"1.6"` instead of `"1.6+"`, causing it to fail on newer JVMs. I will fix it that way, then.

apetresc:
I understand how the version specifier works now. Turns out the original bug was just that it was requiring precisely version `"1.6"` instead of `"1.6+"`. Fixed and works now (whereas the version currently in master does not).

apetresc:
Tests are failing with:
```
==> /usr/local/Cellar/go-gui/1.4.9/bin/gogui -version
No Java runtime present, requesting install.
Error: homebrew/games/go-gui: failed
<0> expected but was
<97>.
```

Not sure why brewbot can't find a Java runtime, especially since the dependency is back in...

apjanke:
Might have been something messed up on the test bot. Build logs have rolled off, so we need to re-run the test anyway.

@BrewTestBot test this please.

apjanke:
Same error during the test.

``` 
       Testing homebrew/games/go-gui
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160909-66142-1a9shca.sb /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/test.rb /usr/local/Library/Taps/homebrew/homebrew-games/go-gui.rb --verbose
==> /usr/local/Cellar/go-gui/1.4.9/bin/gogui -version
No Java runtime present, requesting install.
Error: homebrew/games/go-gui: failed
<0> expected but was
<97>.
```

I can reproduce it locally on my OS X 10.11.6 machine with Java 7 and 8 installed. It looks like what's happening is `go-gui` itself is searching specifically for Java 6, instead of using later versions if they're installed.

The formula is building the `gogui.app` variant of the program. In `go-gui`'s `build.xml`, here's how it's built.

```
<target name="gogui.app" depends="build"
 description="Make Mac application bundle">
<property name="appdir" value="build/GoGui.app"/>
<mkdir dir="${appdir}"/>
<mkdir dir="${appdir}/Contents"/>
<mkdir dir="${appdir}/Contents/bin"/>
<mkdir dir="${appdir}/Contents/MacOS"/>
<mkdir dir="${appdir}/Contents/Resources"/>
<mkdir dir="${appdir}/Contents/Resources/Java"/>
<copy file="/System/Library/Frameworks/JavaVM.framework/Versions/Current/Resources/MacOS/JavaApplicationStub"
  todir="${appdir}/Contents/MacOS"/>
<exec executable="chmod">
<arg line="755 ${appdir}/Contents/MacOS/JavaApplicationStub"/>
...
```

That `JavaApplicationStub` looks like it's specific to Apple's Java 1.6 distribution (in the JavaVM.framework style), and won't pick up the Java 1.7 or 1.8 distributions from Oracle. That might need to be reported and fixed upstream.

On Homebrew's side, we may be able to fix it by switching to building a "normal" Java application (using their Unix build instructions), with the `gogui.jar` target or something similar, and using the normal JAR wrapper script mechanism on it so it launches using the `java` command instead of that `JavaApplicationStub`. I don't think this needs to be an application bundle for Homebrew's purposes.

apjanke:
Yeah, it looks like we could switch to using their `install.sh` script which is intended for Linux but should work for OS X too, if we just tweaked it to remove the Gnome command dependencies.

Anyway, I'd consider the dependency on specifically 1.6 on OS X to be a bug, and it should be reported upstream. (And the unconditional dependency on Gnome commands in `install.sh` seems like a bug, too; not every Unix system is going to have Gnome installed.)

Or for a quick fix, we could just abandon this PR and go back to requiring Java 1.6. Because while `go-gui` on the whole does not require Java 1.6 specifically, the way they're packaging their Mac-specific application bundle apparently does.

I think it'd be better to get this running under any 1.6+ version of Java, though. Lots of people won't have Apple's Java 1.6 installed, and that's an inconvenient requirement since it requires manual installation.

apjanke:
I've put together a PR for switching to a Unix-style installation, which removes the dependency on specifically Java 1.6. Would you mind testing it out? It launches fine for me in my local testing (10.11.6 with Java 1.7 and no Java 1.6), but I'm not a go-gui user so I don't know how to exercise it.

#653

apetresc:
Hey @apjanke – thank you so much for the in-depth investigation :) There is one thing I'm not sure is correct – the `JavaApplicationStub` mechanic isn't specific to Apple's 1.6 distribution – their 1.8 distribution has it as well:

```
 $ /System/Library/Frameworks/JavaVM.framework/Versions/A/Commands/java -version
java version "1.8.0_92"
Java(TM) SE Runtime Environment (build 1.8.0_92-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.92-b14, mixed mode)
$ ls /System/Library/Frameworks/JavaVM.framework/Versions/A/Resources/MacOS/JavaApplicationStub
/System/Library/Frameworks/JavaVM.framework/Versions/A/Resources/MacOS/JavaApplicationStub
```

So it seems the only requirement is that *some* Apple Java distribution is installed, which, I agree, is suboptimal since most people use the Oracle distribution.

The ability to bundle as a `.app` is valuable though; in particular, it's what allows you to associate `.sgf` files with `go-gui` which is an important use case. So I think it's worth investigating the best way to simulate/reimplement `JavaApplicationStub` functionality with the Oracle JVM. I can look into it - do you have any ideas on where to start?

apjanke:
You're welcome!

I'm not sure that's correct about `JavaApplicationStub` working with other JVMs. There's no such thing as an Apple Java 1.7 or 1.8 distribution: they stopped releasing their own Java after 1.6, and Oracle took over, and their releases have a different format.

The OS X command-line `java` command is a wrapper of sorts that uses `/usr/libexec/java_home` or its underlying mechanisms to pick up different actual JVM `java` programs. It doesn't seem like `JavaApplicationStub` does the same thing. That command you ran looks like it's running the `java` from Apple's 1.6 stub framework, and that in turn is using the JVM lookup logic (defaulting to newer versions) to locate an Oracle 1.8 JVM which is also installed on your machine. (Java invocation is a little weird on OS X.)

Here's a third-party 1.7/1.8-aware `JavaApplicationStub` reimplementation: https://github.com/tofi86/universalJavaApplicationStub. Their documentation agrees that the Apple-supplied `JavaApplicationStub` only finds the 1.6 JVMs. Their README looks like a good place to start looking in to this.

> The ability to bundle as a .app is valuable though; in particular, it's what allows you to associate .sgf files with go-gui which is an important use case.  ...

Okay, that makes sense, and is a good reason for keeping the app bundle.

Homebrew is just a packager, and changing the app bundle build to run with a different application launcher is a more extensive change than we usually support. (We want the builds we install to be the same as what you'd get with a vanilla, manual install following upstream's build instructions, to the extent possible.) We'd generally ask upstream to fix their build that way. Maybe you could convince the `go-gui` developers to use that third-party `JavaApplicationStub`? I already opened an issue with upstream talking about the 1.6 requirement: https://sourceforge.net/p/gogui/bugs/42/

Sounds like for the time being, the thing to do is just stick with the exact `:java => "1.6"` requirement, and drop my option of switching to the Unix style build system.

apjanke:
Aha! This looks like a good resource for how to build more modern app bundles from Java applications, replacing `JavaApplicationStub` with `JavaAppLauncher`.

http://dclunie.blogspot.com/2014/10/keeping-up-with-mac-java-bundling-into.html

apjanke:
Mind if I close this PR out, since I think we've established that it's not the right approach?

apetresc:
@apjanke Yeah, I agree – the correct way to solve this is to update it to use a modern app bundler, either upstream or (if not possible) by patching the `build.xml` in the formula. Either way, let's do this in a separate PR once I have time to implement it properly and talk to upstream about it :)

Thanks again for all your help on this!

PR #601, opened 2016-07-22
Replaces location of Rogue tarball file

Davidslv:
Seems that the original website (http://rogue.rogueforge.net/) is no longer available,
I'm replacing it so people can continue installing and playing Rogue.

### DIFF
--- a/rogue.rb
+++ b/rogue.rb
@@ -1,7 +1,7 @@
 class Rogue < Formula
   desc "Dungeon crawling video game"
-  homepage "http://rogue.rogueforge.net/"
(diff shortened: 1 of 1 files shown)


dunn:
Is the project still being maintained?

apjanke:
It's abandonware. But Rogue is of significant historical interest and the build still works. I say we keep the formula alive and go with this URL update. Maybe make https://en.wikipedia.org/wiki/Rogue_(video_game) the homepage.

apjanke:
Merged! Thank you for your contribution to Homebrew.

The formula will install now. If we want to drop it, we can boneyard it in a separate PR.

PR #609, opened 2016-07-31
onscripter 20160728

tomyun:
A hot fix release for iOS. Probably not much changed for OS X package.


dunn:
Merged as fe2796b31c297609edf36271856c5db9962c3288, thanks again!

PR #610, opened 2016-07-31
mame: add revision for tag update

tomyun:
The checksum has changed since two commits were added after initial tagging of 0.176.


dunn:
Can you add a link to where upstream confirmed it was a re-tagging?

tomyun:
I can't find any upstream ticket specifically explaining re-tagging, but just found two additional commits (mamedev/mame@3fc2464d7c6dd154452b45270d20154ea1a25485, mamedev/mame@14e7367f7e1dd575754f8c59fcf74b956c91e87b) after a usual final commit to mark version bump (mamedev/mame@4887ce18443a51fd2a861da959082ad346ff4ee2). I came across this issue by getting an incorrect hash error when trying to build from source.

dunn:
Could you open an issue upstream just to confirm? Thanks!

tomyun:
Now waiting for an upstream answer.

tomyun:
According to the MAME developers, it seems like the hash change was intended to fix some compilation errors, although we didn't hear directly from the person in charge of the release process.

dunn:
That'll do, thanks! Merged as 3e46db2126b5e38cc82b2e25c7656ca0f9c8ca42.

PR #617, opened 2016-08-10
Add git (the confusingly named glulx interpreter)

vickio:
This PR adds a formula for the glulx interpreter `git`. I added some options to allow selecting between three different glk libraries: `glktermw`, `glkterm`, and `cheapglk`. The default is `glktermw`, which enables unicode characters.

I didn't know what to do about naming since this obviously conflicts with the system command of the same name (as well as the `git` formula in `homebrew-core`). I chose `git-if`, short for "git - interactive fiction", but there is no precident for that abbreviation. This is not related to the `git-game` formula.

`glulxe` can use these same glk libraries, though it currently uses `glkterm` and doesn't support unicode. I'm prepared to submit a separate PR to enable glk library selection for `glulxe`, but I wonder if it would be better to split the libraries out into their own formulas and use them as dependencies for `git-if` and `glulxe` (and eventually other interpreters) instead of duplicating all these resources in each formula?


git-if.rb
```diff
@@ -0,0 +1,78 @@
...
+        glk_include.install "glk.h", "glkstart.h", "gi_blorb.h", "gi_dispa.h", "Make.glktermw"
+      end
+    end
+
+    inreplace "Makefile", "#GLK = glkterm", "GLK = #{glk}"
+    inreplace "Makefile", "GLKINCLUDEDIR = ../$(GLK)", "GLKINCLUDEDIR = #{glk_include}"
+    inreplace "Makefile", "GLKLIBDIR = ../$(GLK)", "GLKLIBDIR = #{glk_lib}"
+    inreplace "Makefile", /^OPTIONS = /, "OPTIONS = -DUSE_MMAP -DUSE_INLINE"
+    # This fixes a compile error with clang
+    inreplace "config.h", "__attribute__((__noreturn__))", "__dead2"
```
dunn: Has this been reported upstream?

git-if.rb
```diff
@@ -0,0 +1,78 @@
...
+      glk = "glkterm"
+      resource("glkterm").stage do
+        system "make"
+        glk_lib.install "libglkterm.a"
+        glk_include.install "glk.h", "glkstart.h", "gi_blorb.h", "gi_dispa.h", "Make.glkterm"
+      end
+    else
+      resource("glktermw").stage do
+        inreplace "gtoption.h", "/* #define LOCAL_NCURSESW */", "#define LOCAL_NCURSESW"
+        inreplace "Makefile", "-lncursesw", "-lncurses"
```
dunn: Have these been reported upstream?

git-if.rb
```diff
@@ -0,0 +1,78 @@
...
+    inreplace "Makefile", /^OPTIONS = /, "OPTIONS = -DUSE_MMAP -DUSE_INLINE"
+    # This fixes a compile error with clang
+    inreplace "config.h", "__attribute__((__noreturn__))", "__dead2"
+
+    system "make"
+    bin.install "git" => "git-if"
+  end
+
+  test do
+    if build.with? "cheapglk"
```
dunn: The CI only tests the default options, so it's not really worth writing a test for anything else.

git-if.rb
vickio: Just reported it here: https://github.com/DavidKinder/Git/issues/7

I removed the comment about the compile error because I'm not sure that it's an issue for clang or what is going on.

git-if.rb
vickio: `glktermw` doesn't have an auto config script, so apparently this is the proper way to configure it. From the readme file:

> There are a few compile-time options. These are defined in gtoption.h. Before you compile, you should go into gtoption.h and make any changes necessary. You may also need to edit some include and library paths in the Makefile.

git-if.rb
vickio: What about for users who run `brew test` after installing with `--with-cheapglk`?

git-if.rb
dunn: OK, thanks!

git-if.rb
mistydemeo: What version of OS X/Xcode are you using to build?

vickio: OS X 10.11.6
Xcode command line tools 7.3.1.0.1.1461711523

mistydemeo: Looks like it was fixed in this commit upstream: https://github.com/DavidKinder/Git/commit/36ac371e8fcbf118c12c3dca08a8dc60a5bac93a

This patch stanza can replace the `inreplace`:

```ruby
  # Fixes an issue caused by the use of noreturn as a macro name
  patch do
    url "https://github.com/DavidKinder/Git/commit/36ac371e8fcbf118c12c3dca08a8dc60a5bac93a.patch"
    sha256 "a88ea0da5d323090156ba85441750221f4c790dc9d096002f241b1969f4b0226"
  end
```

vickio: Yep, that does fix it. Added the patch and removed the `inreplace` for `config.h`.

mistydemeo: Thanks!

git-if.rb
```diff
@@ -0,0 +1,82 @@
...
+  # http://www.eblong.com/zarf/glk/index.html
+  resource "glkterm" do
+    url "http://www.eblong.com/zarf/glk/glkterm-104.tar.gz"
+    version "1.0.4"
+    sha256 "473d6ef74defdacade2ef0c3f26644383e8f73b4f1b348e37a9bb669a94d927e"
+  end
+
+  resource "glktermw" do
+    url "http://www.eblong.com/zarf/glk/glktermw-104.tar.gz"
+    version "1.0.4"
```
mistydemeo: Resources within formulae don't need versions; they're not versioned independently of the main formula.

ilovezfs: This is not actually true: https://github.com/Homebrew/brew/issues/517

apjanke:
Looks like the comments have been sufficiently addressed. Sorry for the delay in getting this merged.

The build has aged off the server. Need to re-run it.

@BrewTestBot test this please

apjanke:
BTW,

> ... I wonder if it would be better to split the libraries out into their own formulas and use them as dependencies for git-if and glulxe (and eventually other interpreters) instead of duplicating all these resources in each formula?

Yes, unless the programs using them have very version-specific dependencies and won't necessarily work with the latest version of the libraries. If you want to add support for them to `glulxe`, I'd split them out into their own formulae unless there are version compatibility concerns.

apjanke:

PR #630, opened 2016-08-21
rlvm: revision for glew

tomyun:
- Revision for missing glew library error (Homebrew/homebrew-core#3269)
- Add a patch for missing system header error (eglaysher/rlvm#76)


dunn:
@BrewTestBot test this please

PR #631, opened 2016-08-21
supertux: revision for glew

tomyun:
Fix missing glew library error: Homebrew/homebrew-core#3269


dunn:
@BrewTestBot test this please

PR #633, opened 2016-08-22
openclonk 7.0

tomyun:
The current 6.1 bottle fails to run due to `glew` update (Homebrew/homebrew-core#3269).


tomyun:
Seems to require some C++14 features missing on Mavericks.

```
-- Performing Test HAVE_MAKE_UNIQUE - Failed
CMake Error at cmake/RequireCXXSourceCompiles.cmake:19 (MESSAGE):
  HAVE_MAKE_UNIQUE is required for this project to build properly.  Aborting.
Call Stack (most recent call first):
  CMakeLists.txt:148 (REQUIRE_CXX_SOURCE_COMPILES)
```

Can we just add `depends_on :macos => :yosemite` to drop Mavericks support?

apjanke:
You might also do a `fails_with :clang if MacOS.version <= :yosemite`, which could allow it to compile on Mavericks with gcc. That'd be nicer than dropping Mavericks entirely, IMHO.

tomyun:
I guess I tried that earlier and failed. As I can't believe my own memory, I'll give it another shot here.

tomyun:
There comes another error with `fails_with :clang`:

```
CMake Error at cmake/RequireCXXSourceCompiles.cmake:19 (MESSAGE):
  HAVE_C11_SMART_PTRS is required for this project to build properly.
  Aborting.
Call Stack (most recent call first):
  CMakeLists.txt:146 (REQUIRE_CXX_SOURCE_COMPILES)
```

I fall back to `depends_on :macos => :yosemite` in this case.

apjanke:
That makes sense.

Merged! Thank you for your contribution to Homebrew.

PR #635, opened 2016-08-22
opensyobon: revision for sdl_gfx

tomyun:


dunn:
@BrewTestBot test this please

PR #636, opened 2016-08-22
vecx: revision for sdl_gfx

tomyun:
Existing bottles are probably broken after Homebrew/legacy-homebrew#50411. It's good time to build an El Capitan bottle anyways.


dunn:
@BrewTestBot test this please

PR #638, opened 2016-08-23
libspectrum 1.2.2

tomyun:


apjanke:
@BrewTestBot test this please

apjanke:
This should include a revision bump for fuse-emulator, to get it to link to the updated libspectrum. You can add a `revision 1` right below the `url` and `sha256` lines in `fuse-emulator.rb`.

tomyun:
That's a good call. `revision` added.

tomyun:
Oh I have to make a `revision` in `fuse-emulator`, not here. Reverted. I'll make a revision there once this gets merged.

apjanke:
Actually, the way to do it is to make a `revision` in `fuse-emulator`, but as part of this PR here, so the rebuilt `fuse-emulator` gets tested and deployed as a unit with the new `libspectrum`. You can include revision bumps for any number of dependent formula in a single PR. As far as style goes, each revision bump should be a separate commit. Have a look at https://github.com/Homebrew/homebrew-core/pull/4341 for an example.

apjanke:
Merged in #655 (with revision bump added).

Thank you for your contribution to Homebrew!

PR #648, opened 2016-09-04
fizmo 0.8.2

tomyun:
- Version bump
- Update homepage URLs
- Remove explicit configuration for `ncursesw` (chrender/fizmo-ncursesw@79f2dd9432bfe082f3bfc48a8d3d4e40ead50221)
- Remove version string check (each front-end may have different versions than `libfizmo` which no longer shown up in help message)


dunn:
Thanks again!

PR #616, opened 2016-08-07
blastem 0.4.1

tomyun:
A new version just came out while waiting for #568..

PR #511, opened 2016-02-25
simutrans 120.1.3 (new formula)

tomyun:
[Simutrans](http://www.simutrans.com/) is an open-source transport simulator.

PR #650, opened 2016-09-09
Pachi 11.00 (new formula)

apetresc:
Pachi is one of the strongest open-source Go engines available. It also integrates very well with existing homebrew-games formulas like `go-gui` and `gnugo`.

Passes `brew audit --new-formula`.

### DIFF
--- a/pachi.rb
+++ b/pachi.rb
@@ -3,7 +3,7 @@ class Pachi < Formula
   homepage "http://pachi.or.cz/"
   url "http://repo.or.cz/w/pachi.git/snapshot/pachi-11.00-retsugen.tar.gz"
   sha256 "2aaf9aba098d816d20950d283c8eaed522f3fa71f68390a4c384c0c1ab03cd6f"
-  head "https://github.com/pasky/pachi"
+  head "https://github.com/pasky/pachi.git"
 
   fails_with :clang if MacOS.version <= :mavericks
 
--- a/pachi.rb
+++ b/pachi.rb
@@ -5,6 +5,8 @@ class Pachi < Formula
   sha256 "2aaf9aba098d816d20950d283c8eaed522f3fa71f68390a4c384c0c1ab03cd6f"
   head "https://github.com/pasky/pachi"
 
+  fails_with :clang if MacOS.version <= :mavericks
+
   option "without-patterns", "Don't download pattern files for improved performance"
   option "without-book", "Don't download a fuseki opening book"
 
--- a/pachi.rb
+++ b/pachi.rb
@@ -1,0 +1,51 @@
+class Pachi < Formula
+  desc "pachi"
+  homepage "http://pachi.or.cz/"
+  url "http://repo.or.cz/w/pachi.git/snapshot/pachi-11.00-retsugen.tar.gz"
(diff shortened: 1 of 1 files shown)


tomyun:
Build failed on [Mavericks](https://bot.brew.sh/job/Homebrew%20Games%20Pull%20Requests/993/):

```
==> make
[CC]   board.c
[CC]   gtp.c
[CC]   move.c
[MAKE all]   random
[CC]   random.c
clang: error: unknown argument: '-frename-registers'
clang: error: unknown argument: '-frename-registers'
make: *** [move.o] Error 1
make: *** Waiting for unfinished jobs....
make: *** [board.o] Error 1
clang: error: unknown argument: '-frename-registers'
make: *** [gtp.o] Error 1
clang: error: unknown argument: '-frename-registers'
make[1]: *** [random.o] Error 1
make: *** [all-recursive] Error 1
```

It might be due to outdated C++11 support from `clang` in Mavericks. Do you know of any upstream issues on that?

apjanke:
There's no upstream fixes for older OSes coming.

You could try simply doing a `fails_with :clang` to get it to build with gcc here, possibly conditioning that on the OS version.

You might consider making `patterns` and `book` enabled by default, so users installing from bottles get them.

The fact that you have to set the working directory to take advantage of the downloaded files is pretty lame. That sounds worthy of an upstream bug report.

apetresc:
Thanks for taking a look, all :)

> There's no upstream fixes for older OSes coming.
I agree with that assessment. I'll modify the formula to fall back to `gcc` on Mavericks and lower.

> You might consider making patterns and book enabled by default, so users installing from bottles get them.
Okay, cool. Will do.

> The fact that you have to set the working directory to take advantage of the downloaded files is pretty lame. That sounds worthy of an upstream bug report.
Agreed. In fact, I already have a fix for it in a series of upstream patches I'm preparing, but they might take a bit of time to make it in, so I'll just update the formula when it happens.

apetresc:
The proposed changes have all been made :)

apjanke:
And the test bot is happy. Merged. Thank you for your contribution to Homebrew!

tomyun:
Oh, I was about to make some questions/suggestions for this formula, but maybe it's a bit late. I'll list them anyways:

1. Is there any specific reason to use `share`(`#{prefix}/share`) instead of `pkgshare`(`#{prefix}/share/pachi`)? The latter will be preferred in most cases.
2. You may want to consider having executable colocated with the data path and instead writing a script for running it. `write_exec_script` comes handy in this case: #511, #616, #369.
3. A single squashed commit is preferred for pull. You could try `git rebase -i` and `git push -f`.

apjanke:
It's merged now, but those suggestions would make a good separate PR.

1. Pointing in to `opt` to avoid a specific version dependency may be even better.

apetresc:
@tomyun: Those are great suggestions. I will file those separately later tonight.

As for the squashing, I figured that wasn't needed anymore now that GitHub's merge button has that handy squash option, but I guess you guys don't actually use the button, so I'll make sure to pre-squash next time :)

apjanke:
I squashed the commits myself when merging them this time. (We use a merge tool that allows this.) We do appreciate it if they're already squashed in the PR; saves a bit of work and makes reading the PR easier!

PR #653, opened 2016-09-09
go-gui: Switch to Unix-style installation

apjanke:
This removes the dependency on Apple's Java 1.6 that gogui.app has.

Closes #598.

This change loses the ability to do `brew linkapps go-gui`, because it no longer creates an application bundle. I don't know if that's something `go-gui` users are relying on.

Needs a bit of testing first, since it substantially changes the installation layout.

#### Upstream bugs

https://sourceforge.net/p/gogui/bugs/42/ – Apple Java 1.6 dependency
https://sourceforge.net/p/gogui/bugs/43/ – suggest conditionalizing installation of the Gnome stuff, to eliminate error message and extraneous files


apjanke:
Not going to pursue this. Per [this comment](https://github.com/Homebrew/homebrew-games/pull/598#issuecomment-246113338), the app bundle is indeed useful.

> The ability to bundle as a .app is valuable though; in particular, it's what allows you to associate .sgf files with go-gui which is an important use case.

PR #663, opened 2016-09-14
cataclysm: Add option to save user generated data in home dir

sharpner:
Currently reinstalling the HEAD version will prune
templates, configs and save files. Allowing
users to store those files in $HOME will prevent this.


apjanke:
Perhaps this should be the default, or only option? User data is generally better kept in the user home directory, especially under Homebrew's sandboxed model. Seems to me like this should be the default, unless there are cataclysm-specific compatibility issues, in which case we should probably introduce a `pkgshare`-based data directory.

sharpner:
I also think it should've been the default. But changing it silently could confuse users so I didn't do it that way... what's the policy in brew about BC breaks?

apjanke:
It's really a judgment call. Depends on how many existing users we think there are of the Homebrew package, and the extent of the breakage. Hard to say here.

@mistydemeo, you're the author on many of the previous `cataclysm` commits. What do you think?

sharpner:
Since most people will currently lose their saves on an update anyways, I'll update the PR to use $HOME as default.

apjanke:
That makes sense.

sharpner:
updated :)

zanderzhng:
this should be merged.

sharpner:
Oh I just saw that the tests fail... :/ will need another update *sigh*

zanderzhng:
It works with this commit, I don't know why it failed.

tomyun:
@BrewTestBot test this please

cataclysm.rb
```diff
@@ -42,6 +44,7 @@ def install
 
     args << "TILES=1" if build.with? "tiles"
     args << "CLANG=1" if ENV.compiler == :clang
+    args << "USE_HOME_DIR=1" if ! build.without? "home-directory"
```
tomyun: According to the audit, `args << "USE_HOME_DIR=1" if build.with? "home-directory"` should be enough.

tomyun:
```
Error: 1 problem in 1 formula
homebrew/games/cataclysm:
  * Don't negate 'build.with?': use 'build.without?'
```

cataclysm.rb
sharpner: yeah that's what I had at first, but that would make the "home-directory" setting optional, and not the default. My problem is that I deleted the fork xD

zanderzhng: `args << "USE_HOME_DIR=1" if build.without? "without-home-directory"`

apjanke: tomyun's right, you want `args << "USE_HOME_DIR=1" if build.with? "home-directory"`. That form does not make it the default; the default depends entirely on whether it's written as `option "with-home-directory"` or `option "without-home-directory"` up above.

mistydemeo:
Ping

mistydemeo:
Closing due to lack of response, but please feel free to reopen!

PR #670, opened 2016-09-20
Glk library variations

vickio:
This splits out the 3 Glk library resources from #617 into their own formulae and modifies `glulxe` and `git-if` to build using them.


apjanke:
Looks good to me. I think this is a more maintainable design.

Anyone have concerns before I pull this?

apjanke:

Issue #623, opened 2016-08-13
Bottles files missing


gaoyifan:
The following bottles files return 404:
```
http://homebrew.bintray.com/bottles-games/myman-2009-10.el_capitan.bottle.tar.gz
http://homebrew.bintray.com/bottles-games/myman-2009-10.yosemite.bottle.tar.gz
http://homebrew.bintray.com/bottles-games/myman-2009-10.mavericks.bottle.tar.gz
http://homebrew.bintray.com/bottles-games/tty-solitaire-0.2.0.yosemite.bottle.tar.gz
http://homebrew.bintray.com/bottles-games/tty-solitaire-0.2.0.mavericks.bottle.tar.gz
http://homebrew.bintray.com/bottles-games/tty-solitaire-0.2.0.mountain_lion.bottle.tar.gz
```
Could someone fix them?

dunn:
tty-solitaire bottle updated to 0.2.1 in 32e26fd6838218a69830003d90fcb67667fa8372

tomyun:
myman bottle updated in 21c0acb53e13c868f52ba6400b5fd78b8cb317d5

PR #675, opened 2016-09-21
myman: add head

tomyun:
- add `head`
- primarily for publishing missing bottles (#623)
- `groff` looks like not mandatory at least on Sierra, but leave alone for now


apjanke:

Issue #680, opened 2016-09-25
Update Supertux to version 0.5.0 (stable)


tobbi:
SuperTux 0.5.0, released on September 25. Download from https://github.com/SuperTux/supertux/releases/tag/v0.5.0

tomyun:
Just opened a PR for the new version: #681

It was a bit tricky to have `data` dir correctly recognized by the executable and ended up with modifying `CMakeLists.txt` a bit. Please take a look and let me know if you have any comments.

tobbi:
I am not good with CMake so I have created an issue report here: https://github.com/SuperTux/supertux/issues/614 so that someone can take a look at it.

tomyun:
`supertux` 0.5.0 is now available for install. Thanks for your notification and fix!

PR #681, opened 2016-09-26
supertux 0.5.0

tomyun:
In regards with #680,

- version bump
- use git repo instead of archive (needs `data` dir and some `external` deps)
- remove `physfs` dep (latest version provided by git submodule)
- use standard unix layout instead of .app bundle
- add `inreplace` to set up default data dir correctly
- remove unnecessary files / dirs


supertux.rb
```diff
@@ -1,9 +1,11 @@
...
   desc "Classic 2D jump'n run sidescroller game"
-  homepage "http://supertuxproject.org/"
-  url "https://github.com/SuperTux/supertux/releases/download/v0.4.0/supertux-0.4.0.tar.bz2"
-  sha256 "d18dde3c415e619b4bb035e694ffc384be16576250c9df16929d9ec38daff782"
-  revision 1
+  homepage "https://supertuxproject.org/"
+  url "https://github.com/SuperTux/supertux.git",
+      tag: "v0.5.0",
+      revision: "1531f2c00cf50bad86634cd8e83108579c1d1ca1"
+  sha256 "5a49e2793cc1136f467b5e49ac87e435bfc1438e2a531c567dc2720e88768e65"
```
dunn: If it's a git checkout we don't need a SHA.

supertux.rb
```diff
@@ -20,17 +22,27 @@ class Supertux < Formula
...
+    inreplace "CMakeLists.txt", /SET\((BUILD_DATA_DIR|BUILD_CONFIG_DATA_DIR) .+\)/, "SET(\\1 \"#{pkgshare}\")"
+
+    args = std_cmake_args
+    args << "-DINSTALL_SUBDIR_BIN=bin"
+    args << "-DINSTALL_SUBDIR_SHARE=share/supertux"
+    system "cmake", ".", *args
     system "make", "install"
-    bin.write_exec_script "#{prefix}/SuperTux.app/Contents/MacOS/supertux2"
+
+    # Remove unncessary files
```
tobbi: typo: *unnecessary ;)

supertux.rb
tomyun: Forgot to remove this line. Thanks!

supertux.rb
tomyun: Oops. Now fixed. Thanks!

maths22:
I recommend using the archive provided in the release downloads, rather than doing a git checkout.  It should be significantly faster for the user.  https://github.com/SuperTux/supertux/releases/download/v0.5.0/SuperTux-v0.5.0-Source.tar.gz

supertux.rb
```diff
@@ -20,17 +21,27 @@ class Supertux < Formula
...
   depends_on "glew"
 
   needs :cxx11
 
   def install
     ENV.cxx11
-    system "cmake", ".", "-DDISABLE_CPACK_BUNDLING=ON", *std_cmake_args
+
+    # Correctly set the default data directory in the config.h
+    inreplace "CMakeLists.txt", /SET\((BUILD_DATA_DIR|BUILD_CONFIG_DATA_DIR) .+\)/, "SET(\\1 \"#{pkgshare}\")"
```
maths22: Could you explain the rationale for replacing this line?

tomyun: When first trying to run supertux built out of this formula, it failed, saying something like "`PHYSFS_mount` couldn't add a data directory `/usr/local/share/supertux`: unsupported". Still not sure what caused the problem; maybe physfs doesn't support symlink directory? If this part worked, we would not need replacing here.

Anyways, I looked at the supertux code to find alternative solutions and ended up with modifying `BUILD_DATA_DIR` and `BUILD_CONFIG_DATA_DIR`. Maybe we could also use environment variables, but then a wrapper script would be needed or users had to set their own variables.

tomyun:
@maths22 now uses the archive you mentioned. Somehow confused it with the smaller release archive. Thanks!

maths22:
@tomyun Great!  But now the SHA256 is out of date/wrong :(

tomyun:
@maths22 looks like my cache was messed up. It should now work.

supertux.rb
tomyun: Glad to hear that! So symlink was the culprit?

maths22: Yes (I tried making data a symlink, and that re-created the error on my machine)

tomyun: Instead of replacing CMakeLists.txt directly, your patch (SuperTux/supertux#614) is now incorporated in the formula. I think it's a better way to handle this issue. Also confirmed it's working on my machine.

PR #683, opened 2016-09-28
sdlpop 1.16 (new formula)

tomyun:
- use `var` directory to keep configuration and save/replay files
- executable needs to be run in the same location where those data are stored
- no `test` block (no suitable command for headless testing)


apjanke:
@BrewTestBot test this please

mistydemeo:
@BrewTestBot test this please

mistydemeo:
Not working for me:

```
$ prince
Can't load sprites from resource 1000.
Can't load sprites from resource 700.
```

PR #684, opened 2016-09-28
mame 0.178

tomyun:
- version bump
- build and install the latest `man`
- remove `inreplace` for removing `64` suffix, just use rename install (no need for `if` block as 32-bit support was dropped earlier)


ilovezfs:
rom-tools should be bumped too

tomyun:
Exactly. I didn't have time to look at its changelog yet, but don't expect much change this time. Feel free to open a PR if interested.

ilovezfs:
Cool, I was just making sure you knew they get updated together since I saw the one PR and not the other.

PR #688, opened 2016-09-29
zsdx 1.11.0 (new formula)

tomyun:
[Zelda Mystery of Solarus DX](http://www.solarus-games.org/games/zelda-mystery-of-solarus-dx/) is a Zelda fan game created with the `solarus` engine.
- `SOLARUS_INSTALL_DATADIR` is set to `share`, not `pkgshare`, and it's prepended by the package name.
- `test` actually tests `solarus` engine.

### DIFF
--- a/zsdx.rb
+++ b/zsdx.rb
@@ -1,0 +1,20 @@
+class Zsdx < Formula
+  desc "Zelda Mystery of Solarus DX"
+  homepage "http://www.solarus-games.org/games/zelda-mystery-of-solarus-dx/"
+  url "https://github.com/christopho/zsdx/archive/zsdx-1.11.0.tar.gz"
+  sha256 "05a5d220bbf2439c9da2e71cd9d104240878123fff5bc702e2405d6d0712f0dc"
+  head "https://github.com/christopho/zsdx.git"
+
+  depends_on "cmake" => :build
+  depends_on "solarus"
+
+  def install
+    system "cmake", ".", *std_cmake_args, "-DSOLARUS_INSTALL_DATADIR=#{share}"
+    system "make", "install"
+  end
+
+  test do
+    system Formula["solarus"].bin/"solarus-run", "-help"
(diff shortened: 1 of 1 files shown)


zsdx.rb
```diff
@@ -0,0 +1,19 @@
...
+  depends_on "cmake" => :build
+  depends_on "solarus"
+
+  def install
+    system "cmake", ".", *std_cmake_args, "-DSOLARUS_INSTALL_DATADIR=#{share}"
+    system "make", "install"
+  end
+
+  test do
+    system "#{Formula["solarus"].bin}/solarus-run", "-help"
```
apjanke: This test looks like it's only exercising `solarus`, and not `zsdx`. Can something be done to test the `zsdx` installation itself?

tomyun: Something like `assert File.exist? "#{pkgshare}/data.solarus"` could be possible, if it's a still acceptable test. It seems we don't carry this kind of tests anymore though.

tomyun:
Added a simple test to check if the data files are archived to a valid zip file. It's pretty naive, but maybe better than checking a mere existence.

zsdx.rb
apjanke: Even something trivial like file existence checks would be fine here, and better than checking a different formula's functionality. For GUI programs, we're lenient about allowing trivial tests.

If there's something like a `--version` or `--info` that could be run without invoking the GUI, that would be nice too.

tomyun: `zsdx` itself is more like a collection of scripts and resources to be run on `solarus`. As it does not expose any executable options, I had to instead test its engine, `solarus`. Now I just added another test to check validity of the archive.

apjanke: Okay, that sounds preferable.

mistydemeo:
@BrewTestBot test this please

mistydemeo:
Thanks!

Issue #691, opened 2016-09-30
linker error when brew installing stone-soup - sqlite problem?


viic:
[viic@sherlock:~]$ brew install stone-soup
==> Installing stone-soup from homebrew/games
==> Downloading https://crawl.develz.org/release/stone_soup-0.18.1.tar.xz
Already downloaded: /Users/viic/Library/Caches/Homebrew/stone-soup-0.18.1.tar.xz
==> make install DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer SDKROOT=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk SDK_VER=10.12 MARCH=x86_64 prefix=/usr/loca
Last 15 lines from /Users/viic/Library/Logs/Homebrew/stone-soup/01.make:
      SQL_DBM::finalise_query(sqlite3_stmt**) in sqldbm.o
      SQL_DBM::try_insert(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in sqldbm.o
      SQL_DBM::remove(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in sqldbm.o
      SQL_DBM::do_query(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*) in sqldbm.o
      SQL_DBM::nextkey() in sqldbm.o
  "_sqlite3_step", referenced from:
      SQL_DBM::try_insert(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in sqldbm.o
      SQL_DBM::remove(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in sqldbm.o
      SQL_DBM::do_query(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*) in sqldbm.o
      SQL_DBM::nextkey() in sqldbm.o
[...]

PR #692, opened 2016-09-30
stone-soup: fix sqlite link error

tomyun:
A kind of hot fix for sqlite linker error on Sierra (#691) by making it build its own static version of sqlite. Seems to work fine at least on my local machine. Yet, we'd better need to understand the root cause of the problem.


apjanke:

PR #698, opened 2016-10-01
reminiscence 0.3.2 (new formula)

tomyun:
[REminiscence](http://cyxdown.free.fr/reminiscence/) is a reimplementation of the engine used in the game [Flashback](https://en.wikipedia.org/wiki/Flashback_(1992_video_game)).


apjanke:

PR #671, opened 2016-09-20
mgba 0.5.0

ilovezfs:
Created with `brew bump-formula-pr`.

PR #700, opened 2016-10-06
mgba 0.5.1

tomyun:
- version bump
- use https `homepage`
- update `inreplace` to fix the issue encountered in #671


tomyun:
Seems we need to wait until `qt5` becomes stable on Sierra?

```
Error: qt5 5.6.1-1 did not build
```

ilovezfs:
indeed. you can throw a max os requirement on it in the meantime if you'd like to merge it now

tomyun:
Now it builds on Sierra after `qt5` got updated.

PR #714, opened 2016-10-26
mame 0.179

tomyun:
- version bump
- remove no longer used deps: `sqlite`, `libuv`


apjanke:

PR #722, opened 2016-11-01
bubbros 1.6.2 (new formula)

tsparber:
This adds "The Bub's Brothers" a python clone of the famous bubble bobble game.
This is my first formula created from scratch, so I would be happy to get some feedback and improve the formula if needed.

The python scripts need an absolute path if symlinked. I did not manage to create an absolute symlink, so the `shim_script` is added. Would it be ok to use `#{prefix}` in this shim script? This would simplify the script, but make it harder (impossible?) to relocate the bottle.

For the failing CI see https://github.com/Homebrew/brew/issues/1419.
(fetching `pygame` fails as `hg` is not installed).
I could also make `pygame` an `:optional` dependency, but it is the only "usable" display/audio driver on macOS (X11 has no audio and java needs security changes).

I'm also no sure about the fomula naming, possible variants:
* bubbros (from the tar.gz)
* bubnbros (from the bin as in the `Makefile`)
* bub-n-bros (as the project name on SF)


tsparber:
For the failing CI see https://github.com/Homebrew/brew/issues/1419.
(fetching `pygame` fails as `hg` is not installed).
I could also make `pygame` an `:optional` dependency, but it is the only "usable" display/audio driver on macOS (X11 fails and java needs security changes).

tomyun:
You may want to read [Python for Formula Authors guide](https://github.com/Homebrew/brew/blob/master/docs/Python-for-Formula-Authors.md) if you haven't already. Homebrew provides some helper functions to handle symlink installation and they might be helpful for this case.

In regards with naming, Homebrew prefer names already used in other packager managers (i.e. ubuntu, debian, macports, pkgsrc, etc.). I'd also personally like names matching with their executables.

bubbros.rb
```diff
@@ -0,0 +1,67 @@
+class Bubbros < Formula
+  desc "The Bub's Brothers: Clone of the famous Bubble Bobble game"
+  homepage "http://bub-n-bros.sourceforge.net"
+  url "https://downloads.sourceforge.net/project/bub-n-bros/bub-n-bros/1.6.2/bubbros-1.6.2.tar.gz"
+  sha256 "0ad8a359c4632071a9c85c2684bae32aa0fa278632c49f092dc4078cfb9858c4"
+
+  option "with-x11", "Enable X11 display driver"
+  option "without-pygame", "Disable PyGame display/sound driver"
```
tomyun: These options might be redundant. Specifying `:recommended` or `:optional` automatically generates descriptions for them.

tomyun:
@BrewTestBot test this please

tsparber:
Thank you for your review. I will update the Formula namings to make them similar to the [debian package](https://packages.debian.org/sid/bubbros). They also apply some [patches](https://sources.debian.net/patches/bubbros/1.6.2-1/). I'll try to incorporate them here as well.

tsparber:
Updated to include the patches and namings from the debian package.

mistydemeo:
@BrewTestBot test this please

mistydemeo:
@MikeMcQuaid How do you want to handle the hg problem in CI here?

tsparber:
The error will go away as soon as https://github.com/Homebrew/homebrew-python/pull/360 is merged.

mistydemeo:
@BrewTestBot test this please

PR #731, opened 2016-11-08
cataclysm: Add optional support for LUA

ddunwoody:
Many mods for Cataclysm require LUA support.  This PR adds `LUA=1` to the build command if `--with-lua` is specified, and updates the requirements accordingly.


tomyun:
Looks good to me. Could you add `revision 1` after `sha256` line so that existing users can get update once it's merged? Also minor nitpicks, Homebrew prefers options and deps ordered alphabetically. `lua` stuff can go before `tiles` stuff I guess.

cataclysm.rb
```diff
@@ -16,6 +16,7 @@ class Cataclysm < Formula
   end
 
   option "with-tiles", "Enable tileset support"
+  option "with-lua", "Enable LUA support"
```
tomyun: This can go before tiles option. Also I think 'Lua' is more official than 'LUA'.

ddunwoody:
Changed "LUA" to "Lua", reordered options, deps and ARGS alphabetically.

@tomyun I wasn't quite sure exactly where you'd like the `revision 1`; I'm not at my Mac so committing blind, but can finish this off later today.

tomyun:
@ddunwoody `revision` can be after `sha256` and before `head` I think. If you're not sure, you can always put it somewhere and let `brew audit --strict` tells you where it likes.

ddunwoody:
@tomyum `revision` added.  Thank you for the tip on `brew audit --strict`; most helpful.

tomyun:
It now passes the test and bottles fine. BTW, it gives some debug messages for missing class_definitions.lua and autoexec.lua on startup. Is it something we'd expect?

```
DEBUG: Lua returned file io error for lua/class_definitions.lua
cannot open lua/class_definitions.lua: No such file or directory
  Press spacebar...
```

And I forgot to ask if we'd like `--with-lua` to be `:recommended` (default) rather than `:optional`. If the most users are expected to use lua-dependent modules, `:recommended` would make more sense. I'd leave it up to your decision.

zanderzhng:
being optional should be better. lua support is currently very limited.

tomyun:
@zanderzhng Thanks for your comment. I'll pass on `:recommended` for `--with-lua` option then. Do you have any idea on `lua/class_definitions.lua` error above?

zanderzhng:
i just built cataclysm `--with-lua --HEAD` without error, did you build with HEAD?
it seems to be a new feature, maybe a dependency for HEAD is necessary.

ddunwoody:
I'll update this over the weekend with the recommendation and take a look at the missing file error.

The latter seems to be explicitly referenced in the Makefile.

mistydemeo:
Ping?

mistydemeo:
Closing due to lack of response, but please feel free to reopen!

PR #734, opened 2016-11-11
Polyglot 2.0.4 (new formula)

apetresc:
Polyglot is an adapter for UCI chess engines (i.e, most of them, including [stockfish](https://github.com/Homebrew/homebrew-games/blob/master/stockfish.rb)) and the popular Xboard/Winboard GUI (which homebrew-games [also packages](https://github.com/Homebrew/homebrew-games/blob/master/xboard.rb).) It is **necessary** in order to use stockfish together with xboard. As such, I've also added it as a `:recommended` dependency on the `xboard` formula.

Passes `brew audit --new-formula`.


tomyun:
It looks like the tgz snapshot keeps its hash value changing whenever it gets re-downloaded. Do you think if we can find a more stable archive link that sticks to a fixed hash?

```
==> Verifying polyglot-1.4.70b.git;a=snapshot;h=5904a294b3b33dfd3ca11e4e48133151787f0d8f;sf=tgz checksum
Error: SHA256 mismatch
Expected: dc427cbda292515f0762d4a8e9616541853c34a8c2a13882fae43736ae7b8d71
Actual: 4d86fe411c7bc7cd58dd0b2b786f4564defff756ba4ac2562654e307d634502e
```

apetresc:
Thanks @tomyun - yes, I also just noticed this :( I'm talking to the maintainer now to see what we can do. I'll update the ticket soon!

apetresc:
@tomyun, okay, I have communicated with the project maintainer, and he has provided a stable tarball of the latest release (and also informed me that `master` is not actually the mainline head branch for his development, but `learn` is, so I've updated that too).

I've squashed the commit to make for easier merging. Let me know if there's anything else I should do!

apetresc:
(Not sure how to make @BrewTestBot re-test this)

tomyun:
Great to have a stable tarball now! Test bot is automatically triggered whenever something gets pushed. Let's see how it turns out.

apetresc:
I noticed a bug in the test :( I'll re-push the fix as soon as the bot is done with this run.

apetresc:
(I think I have to ping @BrewTestBot since I'm force-pushing a squashed commit... it doesn't seem to be triggering on its own)

apetresc:
@BrewTestBot re-test this

apetresc:
Looks like tests pass, and `brew audit --new-formula polyglot` is also happy :)

tomyun:
It also builds on my local machine and seems to run fine.

1. We'd better update revision of `xboard` since there's a new dependency added. You can simply add `revision 1` after `sha256` line.
2. Seems like the version of `polyglot` jumped to 2.0.4, not 1.4.70b as in the title. If that's right, please change the title accordingly. Thanks!

apetresc:
@tomyun Done and done :)

tomyun:
Merged. Thanks for your contribution!

PR #741, opened 2016-11-17
xboard: add HEAD support

apetresc:
Add support for building from the official Git repo.

Tests and audit seem happy.


xboard.rb
```diff
@@ -4,14 +4,20 @@ class Xboard < Formula
   url "https://ftpmirror.gnu.org/xboard/xboard-4.9.1.tar.gz"
   mirror "https://ftp.gnu.org/gnu/xboard/xboard-4.9.1.tar.gz"
   sha256 "2b2e53e8428ad9b6e8dc8a55b3a5183381911a4dae2c0072fa96296bbb1970d6"
-  revision 1
+  revision 2
```
tomyun: I think we don't need revision update here for just adding HEAD since there'd be no change for the current users.

xboard.rb
```diff
@@ -4,14 +4,20 @@ class Xboard < Formula
...
   bottle do
     sha256 "95495210b949f1acb8f90d7171e8f60681cd9c556722141fef3869167a3f3fd8" => :sierra
     sha256 "69d6e3d9a107028bc38f47fde16ccebe03236032fedee29388363c1ff3e0227f" => :el_capitan
     sha256 "a715d902478cdf0d4d2e2e205d5119159c670f8ab7f2bd90c28afda4d33d7023" => :yosemite
   end
 
+  head do
+    url "git://git.sv.gnu.org/xboard.git"
+    depends_on "automake" => :build
+    depends_on "autoconf" => :build
```
tomyun: Just a nitpick, but it's been said Homebrew now prefers alphabetical ordering of the deps. We'd better follow this convention at least for the new deps.

tomyun:
Just a few comments in the code. Other than that, it looks good!

apetresc:
Addressed both comments and squashed.

tomyun:
Merged. Thanks for your another contribution to Homebrew!

PR #759, opened 2016-12-26
git-if 1.3.5

equal-l2:


git-if.rb
```diff
@@ -1,9 +1,8 @@
 class GitIf < Formula
   desc "Glulx interpreter that is optimized for speed"
   homepage "http://ifarchive.org/indexes/if-archiveXprogrammingXglulxXinterpretersXgit.html"
-  url "http://ifarchive.org/if-archive/programming/glulx/interpreters/git/git-134.zip"
-  version "1.3.4"
-  sha256 "3ef7bd4ffddfe617cf914ae2f46dc28a30b4c5989590f93c97c53b1eb71a36c2"
+  url "https://github.com/DavidKinder/Git/archive/v1.3.5.tar.gz"
```
mistydemeo: Out of curioisy, is there a reason you're using this URL instead of `http://ifarchive.org/if-archive/programming/glulx/interpreters/git/git-135.zip`?

equal-l2: Well, I don't remember why I do such a strange act.
Using `http://ifarchive.org/...` is right.

git-if.rb
```diff
@@ -1,9 +1,8 @@
 class GitIf < Formula
   desc "Glulx interpreter that is optimized for speed"
   homepage "http://ifarchive.org/indexes/if-archiveXprogrammingXglulxXinterpretersXgit.html"
-  url "http://ifarchive.org/if-archive/programming/glulx/interpreters/git/git-134.zip"
-  version "1.3.4"
```
mistydemeo: Still need the `version` specified manually - it gets misparsed as `135` otherwise.

equal-l2: Thanks for information.

equal-l2: Fixed.
Thanks for the information.

mistydemeo:
Thanks!

PR #760, opened 2016-12-26
ski 6.11

equal-l2:


mistydemeo:
Looks like the test is failing?

equal-l2:
The cause is that a termcap `sgr0` is unavailable in the sandbox.

mistydemeo:
Hm, maybe this test isn't reliable enough to run on the build bots.

mistydemeo:
Should we just remove the test?

equal-l2:
@mistydemeo 
I think this is a upstream bug.

This patch will fix the test.

mistydemeo:
Thank you! Mind reporting it upstream?

equal-l2:
@mistydemeo 
I want to do so, but unfortunately the upstream repo (should be accessible from http://www.catb.org/~esr/ski/ ) seems to be gone.

mistydemeo:
Mind sending ESR an email? I don't think he has a tracker but I think that's the main way he accepts contributions.

equal-l2:
I've emailed to ESR.

By the way, will this PR be merged?
Or do I have to wait for next release?

mistydemeo:
Merged, thanks!

PR #761, opened 2016-12-26
wesnoth 1.12.6

equal-l2:
And bumped devel to 1.13.6


mistydemeo:
Looks like the new devel isn't working; it requires sdl2 instead of sdl.

equal-l2:
Hmm, the test seems to be unsuitable for devel.

mistydemeo:
> objc[79720]: Class SDLApplication is implemented in both /usr/local/opt/sdl2/lib/libSDL2-2.0.0.dylib and /usr/local/Cellar/wesnoth/1.13.6/bin/wesnoth. One of the two will be used. Which one is undefined.

That's not good.

wesnoth.rb
```diff
@@ -13,13 +12,33 @@ class Wesnoth < Formula
...
-    url "https://downloads.sourceforge.net/project/wesnoth/wesnoth/wesnoth-1.13.1/wesnoth-1.13.1.tar.bz2"
-    sha256 "4423645f58eae3a22cdb94736a20181fbb3c2de3de36e20a70084b15b20337f2"
+    url "https://downloads.sourceforge.net/project/wesnoth/wesnoth/wesnoth-1.13.6/wesnoth-1.13.6.tar.bz2"
+    sha256 "a10a193ef8bc9963b2e241826628a120bb7dda758eb2d25d9ad110ad4916937c"
   end
 
   option "with-ccache", "Speeds recompilation, convenient for beta testers"
   option "with-debug", "Build with debugging symbols"
 
+  if build.stable?
```
mistydemeo: You can move this into a `stable do` block, along with the `url` and `sha256`, then move the other dependencies into the `devel do` and `head do` blocks.

equal-l2:
>objc[79720]: Class SDLApplication is implemented in both /usr/local/opt/sdl2/lib/libSDL2-2.0.0.dylib and /usr/local/Cellar/wesnoth/1.13.6/bin/wesnoth. One of the two will be used. Which one is undefined.

I've found issue for this problem.
https://gna.org/bugs/index.php?25357

dunn:
Yeah, that was me.  Here's my updated formula, in case it's a useful reference: https://github.com/dunn/homebrew-games/blob/3264ac1603a0c0032aeafc4275f81078e904c37b/wesnoth.rb

wesnoth.rb
```diff
@@ -13,8 +24,22 @@ class Wesnoth < Formula
...
 
   devel do
-    url "https://downloads.sourceforge.net/project/wesnoth/wesnoth/wesnoth-1.13.1/wesnoth-1.13.1.tar.bz2"
-    sha256 "4423645f58eae3a22cdb94736a20181fbb3c2de3de36e20a70084b15b20337f2"
+    url "https://downloads.sourceforge.net/project/wesnoth/wesnoth/wesnoth-1.13.6/wesnoth-1.13.6.tar.bz2"
+    sha256 "a10a193ef8bc9963b2e241826628a120bb7dda758eb2d25d9ad110ad4916937c"
+    depends_on "sdl2"
+    depends_on "sdl2_image" # Must have png support
+    depends_on "sdl2_mixer" => "with-libvorbis" # The music is in .ogg format
+    depends_on "sdl2_net"
```
dunn: This isn't required; see https://raw.githubusercontent.com/wesnoth/wesnoth/1.13.6/changelog

equal-l2:
Hmm, @dunn, you seem to understand wesnoth more than me.
Would you mind to update the formula by yourself instead of me?

dunn:
Well until the developers fix the bug we both discovered, neither of us can update it!

MikeMcQuaid:
This has been moved to homebrew/core so can be reopened there when working.

PR #762, opened 2016-12-26
widelands build19

equal-l2:


equal-l2:
Migration to SDL2 and C++11 is also needed.

widelands.rb
```diff
@@ -16,42 +16,34 @@ class Widelands < Formula
...
-      system "cmake", "..", "-DLUA_INCLUDE_DIR:PATH=#{Formula["lua51"].opt_include}/lua-5.1",
-                      "-DLUA_LIBRARIES:STRING=#{Formula["lua51"].opt_lib}/liblua.5.1.dylib",
-                      "-DLUA_LIBRARY:FILEPATH=#{Formula["lua51"].opt_lib}/liblua.5.1.dylib",
+  def install
+    ENV.cxx11
+    mkdir "build" do
+      system "cmake", "..",
                       # Without the following option, Cmake intend to use the library of MONO framework.
                       "-DPNG_PNG_INCLUDE_DIR:PATH=#{Formula["libpng"].opt_include}",
-                      "-DWL_INSTALL_DATADIR=#{pkgshare}/data", "-DWL_INSTALL_BINDIR=#{libexec}", *std_cmake_args
```
mistydemeo: Can we continue installing the data into `pkgshare`?

widelands.rb
```diff
@@ -16,42 +16,35 @@ class Widelands < Formula
...
+                      "-DWL_INSTALL_DATADIR=#{pkgshare}/data",
+                       *std_cmake_args
       system "make", "install"
+
       (bin/"widelands").write <<-EOS.undent
         #!/bin/sh
-        exec #{libexec}/widelands --datadir=#{pkgshare}/data "$@"
+        exec #{prefix}/widelands "$@"
```
mistydemeo: Probably still need to pass the `datadir`, right?

equal-l2: I don't think so.
`WL_INSTALL_DATADIR` will be hardcoded in the binary, and the binary will look the hardcoded path wherever the binary is.

mistydemeo: Excellent. That sounds like it should handle it. Thanks very much for checking!

mistydemeo:
Thanks!

PR #766, opened 2016-12-27
wumpus: fix patch locations after move to Gitlab

rwhogg:
ESR moved `wumpus` to Gitlab and the patch links no longer work (404). However, the patches are still in Git, so grab them from Gitlab instead.


mistydemeo:
Thanks!

PR #767, opened 2016-12-27
supertux: Re-add patch

equal-l2:
The patch was supposed to be included in the latest release.


equal-l2:
Also some tweaks added.

mistydemeo:
Should we add a revision to make sure that anyone who got the version without the patch gets this version bump?

equal-l2:
@mistydemeo 

> Should we add a revision

You're right.
I've added revision now.

mistydemeo:
Thanks!

PR #768, opened 2016-12-28
--HEAD support for arx-libertatis

poldy:
I have an open pull req with upstream to fix the build on Mac OS X. Assuming this is accepted, will need to install --HEAD until the next release is cut.

### DIFF
--- a/arx-libertatis.rb
+++ b/arx-libertatis.rb
@@ -12,7 +12,7 @@ class ArxLibertatis < Formula
   end
 
   head do
-    url "https://github.com/poldy/ArxLibertatis.git"
+    url "https://github.com/arx/ArxLibertatis.git"
 
     depends_on "cmake" => :build
     depends_on "inkscape" => :build
--- a/arx-libertatis.rb
+++ b/arx-libertatis.rb
@@ -11,6 +11,14 @@ class ArxLibertatis < Formula
     sha256 "8f1938381423b45d798c8a2665484b4f9a09b8070443b80986e935d07b0410af" => :mavericks
   end
 
+  head do
+    url "https://github.com/poldy/ArxLibertatis.git"
+
+    depends_on "cmake" => :build
+    depends_on "inkscape" => :build
+    depends_on "imagemagick" => :build
+  end
+
   option "without-innoextract", "Build without arx-install-data"
 
   depends_on "cmake" => :build


arx-libertatis.rb
```diff
@@ -11,6 +11,14 @@ class ArxLibertatis < Formula
     sha256 "8f1938381423b45d798c8a2665484b4f9a09b8070443b80986e935d07b0410af" => :mavericks
   end
 
+  head do
+    url "https://github.com/arx/ArxLibertatis.git"
+
+    depends_on "cmake" => :build
+    depends_on "inkscape" => :build
```
equal-l2: `Inkscape` is not in homebrew-core but in homebrew-gui.
You may want to replace this with `homebrew/gui/inkscape`.

equal-l2:
The test fails because of the stable version (1.1.2).

This patch may be a help. (auto-generated from a commit by Github)
https://github.com/arx/ArxLibertatis/commit/442ba4af978160abd3856a9daec38f5b6e213cb4.patch

arx-libertatis.rb
dscharrer: You can avoid the inkscape and imagemagick dependencies for git builds by pulling the pre-built icons from the [ArxLibertatisData](https://github.com/arx/ArxLibertatisData/) repo as described in the [README](https://github.com/arx/ArxLibertatis#git-build-dependencies). (For release version these will be included in the source download.).

Also, I'm not familiar with Homebrew build scripts, but it looks like the repeated cmake dependency here is redundant - either that or you will also need to duplicate the other dependencies.

poldy:
Ah, I see. The buildbot only runs "brew install arx-libertatis". This will fail until upstream cut their next release. My pull req only fixed installing --HEAD for now.

I think it's best to be patient and wait. Until then, anyone hitting the problem can find this pull req.

mistydemeo:
How big is your PR? I don't know how much `HEAD` has changed since the current release; do you think it's feasible to backport the fix as a patch to the stable version?

mistydemeo:
Ping

poldy:
Sorry for the delay, I was travelling. My contribution was only minor, the bulk of the work was done upstream, e.g. defect [#918](https://bugs.arx-libertatis.org/arx/issues/918), together with probably many more fixes in their git repo. So I think backporting all the necessary fixes to 1.1.2 might be a lot of work.

I think it's best to wait for a new release from upstream. Anybody determined can apply this patch to their local clone & build from --HEAD.

PR #769, opened 2016-12-29
Delete unused formulae

MikeMcQuaid:
Delete all homebrew/games formulae that show no installations since analytics records began.

CC @Homebrew/games


mistydemeo:
For what it's worth, I have several of these installed. Some of them in the list were also only added within the last few days.

MikeMcQuaid:
@mistydemeo Interesting. Did you use any options? Do you have analytics disabled? If "no" to both: this is likely due to analytics being statistical so very low numbers of installs can equate to no installs. I think the same applies, though: we probably don't want to keep around stuff that's used by a tiny number of people.

Some higher-level related planning: the goal is to fold this tap and most of the other Homebrew/* ones (except e.g. science, php, nginx) back into Homebrew/core.

All that said, I'm happy to remove specific ones from this list that you use or that have been added recently enough you don't think it's fair to remove them.

mistydemeo:
The following were added in the last five days:

bubbros
sdlpop
zsdx
zsxd

I use the following:

robotfindskitten
slashem
ucon64

I also notice a number of formulae in this list that regularly get PRs; I don't think the stats are accurately capturing the pattern of use in this tap. They're definitely still niche, but they're niche things that are used by a community that's doing a good job of supporting them. I'm perfectly happy to continue putting in the work to continue supporting these.

MikeMcQuaid:
Can you suggest what may capture the pattern of use in this tap? When these are moved en-masse to Homebrew/core they will need to have green CI to move over so the deletions are trying to reduce the work in fixing up and chasing down errors on those that aren't used (or certainly aren't widely used). This may mean that in the medium-term some of these may be better off in third-party taps.

ilovezfs:
Unlike, say, old versions formulae with zero installs, which have no reasonable prospect of increased usage, I do think there's a chance some of these may gain users with the additional exposure homebrew/core provides. My guess is we have a lot of users who have no idea this tap even exists.

ilovezfs:
I think there may be a bug in the script. I see 12 robotfindskitten in the last 30 days.

mistydemeo:
As @rwhogg says in Homebrew/homebrew-emacs#532, I'd rather see this spun out of the Homebrew org entirely than see 58% of the games packages deleted. I don't see deleting so many packages as a good solution, particularly for a tap whose original purpose was to provide a home for packages which wouldn't meet the inclusion criteria for the core repo.

tomyun:
As an author of some of these unused formulae, I have to admit many of them may look obsolete in a sense that they no longer receive active maintenance or foresee heavy use by end-users. Gaming and/or emulator scenes are already quite small on macOS and what Homebrew could cover with CLI-oriented environment should be even smaller.

Yet I believe this tap has been growing up something like a curated list of (open source) games or emulators otherwise would have been forgotten. Indeed many of their original packages no longer run easily on the latest macOS without proper source compilations which could be daunting tasks for most users. Even in the case of active projects, Homebrew formulae often provide flexibility in configuration and devel/head availability than their official packages.

Of course, whether we need to keep this tap official could be still called into question, and I'd personally like to see continuous support from the core team, especially with its great CI and bottle management, as well as the brand itself. Note that some projects already recommend this tap to users to install their macOS packages (i.e. Freeciv, CorsixTH, EasyRPG, Fuse, to name a few).

By the way, I also found many formulae installed on my machine are listed unused, maybe due to its statistical insignificance.

MikeMcQuaid:
> I don't see deleting so many packages as a good solution, particularly for a tap whose original purpose was to provide a home for packages which wouldn't meet the inclusion criteria for the core repo.

I think the problem we had with taps is that what started as "we won't accept this but we might in another tap" became "it's a game so lives in homebrew/games". I'm definitely interested in migrating a bunch of these formulae into homebrew/core and leaving some of these here but after the tap consolidation is done it's probably worth moving this to e.g. @mistydemeo's personal account.

> Of course, whether we need to keep this tap official could be still called into question, and I'd personally like to see continuous support from the core team, especially with its great CI and bottle management, as well as the brand itself. Note that some projects already recommend this tap to users to install their macOS packages (i.e. Freeciv, CorsixTH, EasyRPG, Fuse, to name a few).

Yep, that's what I'm proposing: those things that are popular be more actively maintained by the core team rather than less maintained. If this works for everyone here I'm happy to just migrate packages across whenever and leave the rest of these as-is and we can consider moving the tap at a later stage.

MikeMcQuaid:
Passing on this until a later migration.

PR #773, opened 2017-01-05
mgba 0.5.2

equal-l2:


equal-l2:
@ilovezfs 
Why did you make the test abort?

ilovezfs:
because it was stuck having not completed for two hours

equal-l2:
Well, the new test is only applicable to `mgba` without Qt.
With Qt, the former test is suitable.

PR #776, opened 2017-01-16
nethack: add HEAD

jonchang:


mistydemeo:
What an era where they have a public source tree!

PR #778, opened 2017-01-23
mednafen 0.9.41

tomyun:


mistydemeo:
Looks like it's not building with GCC 6.

mistydemeo:
I decided to pull with a Sierra bottle only. If people are able to fix builds on older OSs, they can get readded.

PR #783, opened 2017-01-28
mame 0.182

mbcoguno:
* Add option to build MAME's tools inside the MAME formula.


mbcoguno:
I have a question.
As rom-tools isn't an independent build target now, and rarely user download this formula bottle, I think it's time to migrate them into a single formula (MAME), and delete the other one (rom-tools).
Any suggestion?

dunn:
Fine with me! cc @tomyun

mistydemeo:
Given how long the build is, maybe we should enable the tools by default so that users who want them don't have to endure a full MAME compile?

tomyun:
As I already talked to @mbcoguno in email conversation, I'm also fine with this change.

@mistydemeo I think enabling the tools by default should be careful given that we're installing more than a dozen of executables which are irrelevant to most users. Even many of them do not run properly on macOS as we can see on the test cases.

mbcoguno:
On my laptop, build MAME without tools spend 32 minutes 50 seconds, and with tools spend 33 minutes 12 seconds.
BTW, MAME can build with keg_only's lua@5.3.

mbcoguno:
If someone still need rom-tools bottle, change build flag `TARGET=mame SUBTARGET=tiny` in rom-tools formula is enough, it builds 8 minutes on my laptop.
But I think rom-tools will be delete due to cleanup tap's unused formulae, if someone think preserve rom-tools bottle is still necessary, I will edit the commit to preserve rom-tools.

tomyun:
@mbcoguno When I tried tiny subtarget on an earlier version of mame last month, somehow it didn't work out of scratch. It seemed to require some modules built first with a regular target. It'd be a nice way to keep `rom-tools` if it works now.

mbcoguno:
I'm not using rom-tools in depth. I'm just testing them. Since the compiled binary isn't same with regular MAME build (hash value if different between mame and tiny target), and rom-tools isn't working well in macOS, and tools build script isn't stable (in my update formula experience), I think remain this PR's commit (delete rom-tools) is fine. If anyone want rom-tools, build from source with regular build target just enough.
Sorry, I'm hesitate on it before.

tomyun:
@mbcoguno agreed. I don't particularly want to keep `rom-tools` either, but just remember some users were against it when I suggested formulae merge a while ago.

MikeMcQuaid:
This was moved to homebrew/core so if we want to merge this it'll be over there.

PR #787, opened 2017-02-03
ski 6.12

equal-l2:


ilovezfs:
duplicates https://github.com/Homebrew/homebrew-games/pull/786

But thanks for the PR nonetheless @equal-l2

